PR #646, opened 2021-06-03
[DEVEX-1250] Fix code/preview toggle in doc-site for IE11

danidewitt:
## What did we change?
- Fixed the code/preview toggle in the doc-site for IE11.

## Why are we doing this?
- The toggle between preview and code in the doc-site wasn't working in IE11. I was getting an error about resources not found and not rendering react. Read a bunch about this issue people were having, seemed like their issues were related to gatsby cache...not our issue. Narrowed it down to the regex used to strip the `react-router-dom` imports in the playroom examples. Turns out javascript doesn't support regex lookbehinds yet.

## Screenshot(s) / Gif(s):
Screen Shot 2021-06-03 at 10 38 38 AM
Screen Shot 2021-06-03 at 10 38 30 AM

## Checklist

- [ ] Provide test coverage for the changes made
- [ ] Create a changeset (`npm run changeset`) with a summary of the changes

[Contributing guidelines](https://ezcater.github.io/recipe/guides/contributing/)

PR #647, opened 2021-06-04
Fix: code/preview toggle in doc-site for IE11

CraigCav:
## What did we change?

fix: remove regex lookbehind from doc-site for IE11 support

switched code rewriting (via lookbehind regex) to using `scope` to replace the `require` function with a fixed implementation that knows how to import react-router. Replacing "global" variables and methods via scope is consistent with the approach we use in our [visual regression tests](https://github.com/ezcater/recipe/blob/main/src/components/EzPageHeader/__tests__/EzPageHeader.test.tsx#L23-L31) and in [`markdown.md`](https://github.com/ezcater/recipe/blob/5923c30edf13660125a8f19b4495682b13283947/doc-site/src/components/Markdown.js#L84-L92).

## Why are we doing this?

The doc-site is currently broken in IE11, as the regex uses syntax that is not supported in IE.

DEVEX-1250

## Checklist


CraigCav:
Alternative to #646. 

I'm trying to avoid having to lean on regex to re-write code examples. While the regex replacement does work for the current examples we have in our docs today, I'm concerned that it would not work with any small alteration to syntax, or if folks write their own require calls expecting them to work in other examples.

Instead of re-writing the code, I've switched out the implementation of require with a version that handles only react-router imports and throws a specific error if you try to require anything else in an example.

danidewitt (approved):
i pulled it down and can confirm it works with the playroom 🚢

PR #651, opened 2021-06-08
EzModal support for Slots

CraigCav:
## What did we change?

- Added design docs for slots API for EzModal
- Removed `CodePlaygroundHostNode` from EzModal documentation examples
  - This was a hold-over from having either an iframe OR a host page live demo. Now we're not hosting code examples inside an iframe, we no longer need the `CodePlaygroundHostNode`
- refactor: convert EzModal internals to use EzCard for consistent styling
- feat: Add slots API implementation to EzModal
- feat: Add first-class support for EzPreview (with tests) for EzModal with images.

## Why are we doing this?

In order to make EzModal support multiple complex layouts to solve a variety of use cases, while still making it easy place content in the right places regardless of the components used or what order they appear in the DOM.

## Screenshot(s) / Gif(s):

image

image

## Checklist


src/components/EzModal/EzModal.md
```diff
@@ -18,21 +18,63 @@ Modals should:
...
+The `EzHeader`, `EzContent`, and `EzFooter` content elements accept any renderable node, not just strings, allowing you to create dialogs for more complex workflows.
+
+A modal can be shown or hidden using an [`EzDialogTrigger`](/components/ez-dialog-trigger), which will toggle the modal's open state on press of the trigger. Alternatively, the modal can accept an `isOpen` prop to control whether the dialog is shown or hidden. This can be useful in cases where the is no trigger element, or when the trigger would unmount while the modal is open, e.g a confirmation of deletion dialog.
+
+## Example
 
-Informational modals are used to focus the user’s attention on specific information. Use when providing information about something that’s not essential to completing the actions on the page, especially if content formatting is needed or there’s too much content to display in a tooltip.
+```jsx
+<EzDialogTrigger isDismissable>
+  <EzButton>Open</EzButton>
```
ez-alexfrazer, line 47: How does this connect to the `EzDialogTrigger`? Does the `EzDialogTrigger` map/clone the children? 🤔 In theory, can you do something like this?

```tsx
<EzDialogTrigger>
  {close => (
    <EzModal>
      <EzButton onClick={close}>Close</EzButton>
    </EzModal>
  )}
  <EzButton>Open</EzButton>
</EzDialogTrigger>
```

CraigCav: This could work one of a few ways (which route to take isn't finalized).

1.  We standardize the order of the children, so we can provide the wireup of the button via cloning. This would be similar to (but not exactly how) [Adobe's dialog component works](https://react-spectrum.adobe.com/react-spectrum/DialogTrigger.html#content).
2. We leverage the Slots API to provide "slot props" to named slots. The "open" button would receive the slot props for the wireup of the trigger event handler(s) and the modal would receive any props we want to pass down to the modal (such as the `open` state). The modal would internally use our `ClearSlots` component such that it's children (i.e. the close button) don't also receive the trigger slot props. In this setup, the order of the children is not important.

Whichever route we go, a summary of any constraints will be covered in the docs for the `EzDialogTrigger` component.

CraigCav: The value of `EzDialogTrigger` is that it would manage the state for showing and hiding the modal, simplifying the common case of using a trigger element to pop up a modal by removing some of the plumbing code.

In more complex cases (such as multiple triggers for the same dialog, or the dialog living longer than the trigger element), then you'd still be able to drop back to handling this state within the application.

CraigCav: I've pulled the EzDialogTrigger API out of this PR as it isn't needed to bring the Slots API to EzModal.

That'll give us some more time to think some more about what we want to do without holding up slots.

playroom/useScope.js
```diff
@@ -15,5 +14,4 @@ export default () => ({
     // fake out calls to `require('react-router-dom')` within our .md examples.
     return {BrowserRouter, StaticRouter, NavLink, Link, Route};
   },
-  CodePlaygroundHostNode: ({children, openState}) => <div openState={openState}>{children}</div>,
```
danidewitt, line 18: 👍 nice

src/components/EzCard/EzCard.styles.tsx
```diff
@@ -86,13 +86,13 @@ export const msGrid = maxWidth =>
     variants: {
       imagePosition: {
         top: {
-          '-ms-grid-columns': '0px minmax(0px, 100%)',
+          '-ms-grid-columns': '0px minmax(0px, 1fr) min-content 0px',
```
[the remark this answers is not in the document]
danidewitt: TIL `fr`

src/components/EzCard/EzCard.styles.tsx
```diff
@@ -86,13 +86,13 @@ export const msGrid = maxWidth =>
...
         },
         right: {
-          '-ms-grid-columns': `0px minmax(0px, 100%) ${maxWidth || '50%'})`,
+          '-ms-grid-columns': `0px minmax(0px, 1fr) min-content ${maxWidth || '50%'})`,
```
danidewitt, line 92: i assume all these `-ms-grid-columns` additions mean it looks good in IE11 too

src/components/EzModal/EzModal.md
```diff
@@ -108,10 +156,11 @@ The `destructive` prop should be used to indicate that the `onSubmit` function i
 
 ```jsx
 () => {
-  const [isOpen, setIsOpen] = React.useState(true);
```
danidewitt, line 111: i never noticed these were all default `true` before...

src/components/EzModal/en.ts
```diff
@@ -0,0 +1,3 @@
+export default {
+  dismiss: 'Dismiss',
```
danidewitt, line 2: 👍

danidewitt (approved):

PR #654, opened 2021-06-09
[DEVEX-1200] Multiple choice support for nested content

danidewitt:
## What did we change?
- Updated `EzChoice` to allow for nested content
- Added `EzItem`

## Why are we doing this?
- To allow for proper nested content in checkbox and radio fields [DEVEX-1200](https://ezcater.atlassian.net/browse/DEVEX-1200) 

## Screenshot(s) / Gif(s):
Screen Shot 2021-07-16 at 12 16 09 PM

Screen Shot 2021-07-16 at 12 16 18 PM

Screen Shot 2021-07-16 at 11 40 03 AM

## Checklist


src/components/EzField/EzField.md
```diff
@@ -206,6 +206,47 @@ Use the `options` prop to provide an array of options for selection. Each option
...
+      <EzField
+        type="radio"
+        label="Single choice list"
+        value={selectedChoice}
+        onChange={e => setSelectedChoice(e.target.value)}
+      >
+        <EzItem key="a">
+          <EzLabel>Choice A</EzLabel>
+          <EzContent>
+            <EzWell>
```
danidewitt, line 231: we'll need to build the well or remove them from the example

danidewitt: we'll also need to build the ezitem, and handle the case where there aren't options to map over

danidewitt: obviously these things will be taken care of when i build the slots architecture for this field type -- these were more just notes to myself to not merge this PR until we do all of this

src/components/EzField/EzField.md
```diff
@@ -226,15 +226,15 @@ Instead of using the `options` prop to provide an array of options to a radio or
         onChange={e => setSelectedChoice(e.target.value)}
       >
         <EzItem key="a">
-          <EzLabel>Choice A</EzLabel>
+          <EzLabelledItem position="top" size="normal" title="Choice A" />
```
CraigCav, line 229: I understand why this change is currently necessary... but I can't help but think we'd be better off exposing `EzLabel` (or something else) and deprecating `EzLabelledItem`.

There are also a couple of other challenges to consider here:

- `EzLabelledItem` expects you to provide an `htmlFor` prop when labelling input fields (and a radio is a type of input field), but we're not currently providing one and it's not clear how we'd even provide this value.
- It's probably unclear for users what `position` to use. In theory, we might even need `position: right` since the label is to the right of the radio, except that option isn't available on `EzLabelledItem`  today.


I suspect we can lean on "slot props" here to handle some of the wire up (such as `htmlFor`) for us, however we will need to ensure our docs make it clear that this is happening behind the scenes for this component.

danidewitt: yeah, maybe this goes with my comments above about `EzItem` and `EzWell` needing to be built. we'll have to decide what to do with `EzLabel` - sounds like `EzLabel`? maybe it will be more clear when i start wiring stuff up, and/or we can discuss more

src/components/EzField/EzChoice.tsx
```diff
@@ -46,25 +47,47 @@ const box = theme.css({
...
+    optionHasChildren && label.find(child => child.type.displayName === 'EzLabel');
+  const contentFromChildren =
+    optionHasChildren && label.find(child => child.type.displayName === 'EzContent');
   const id = useUniqueId();
   return (
     <Style ruleset={theme}>
       <span className={box({bordered, disabled})}>
         {React.cloneElement(input, {id})}
-        <label htmlFor={id}>{label}</label>
+        {labelFromChildren || <label htmlFor={id}>{label}</label>}
```
danidewitt, line 67: The labels look different, depending on if you provide `options` and get a `<label>` or if you provide children and get a `<EzLabel>`. Thoughts?

`options` and `EzLabel`:
Screen Shot 2021-06-28 at 10 28 04 AM

`children and `label`:
Screen Shot 2021-06-28 at 10 27 59 AM

src/components/index.tsx
```diff
@@ -7,12 +7,13 @@ export {default as EzButton} from './EzButton';
 export {default as EzCarousel} from './EzCarousel';
 export {default as EzCheckbox} from './EzCheckbox';
 export {EzCard, EzCardSection, EzCardExpandable} from './EzCard';
-export {EzContent, EzHeader, EzFooter, EzPreview} from './EzContent';
+export {EzContent, EzHeader, EzItem, EzFooter, EzPreview} from './EzContent';
```
danidewitt, line 10: Do we need to add docs or anything else for `EzItem`?

src/components/index.tsx
```diff
@@ -7,12 +7,13 @@ export {default as EzButton} from './EzButton';
...
 export {default as EzField} from './EzField';
 export {default as EzFlashMessage} from './EzFlashMessage';
 export {default as EzFormLayout} from './EzFormLayout';
 export {default as EzGlobalStyles} from './EzGlobalStyles';
 export {default as EzHeading} from './EzHeading';
+export {default as EzLabel} from './EzLabel';
```
danidewitt, line 16: Do we need to add docs or anything else for `EzLabel`?

src/components/EzField/EzChoice.tsx
CraigCav: I think we still need to give some thought over whether `EzLabel` (in its current form) is something we should expose.

From an API perspective, expecting folks to pass `<EzLabel as="label" size="normal">` is less than ideal; `label` should really be the default, as should `normal`. 

That being said, I think it's a little awkward that to get the intended styles, we'd have to *remove* styles from `EzLabel` (or override them). By contrast, in React Spectrum, "slot" components (like Content, Text, Header, Heading etc) specifically are described as having "no specific styling by itself, but receives styling from the parent container." It might be worth considering making the exported `EzLabel` component a separate component from the internal `EzLabel` we have today for styling labels (i.e. create a new component, and renaming the existing one as `StyledLabel` or something).

CraigCav: Another consideration - how do we wire up the `htmlFor` attribute to the provided `EzLabel` component? It needs to be wired up to ensure that clicking the label (instead if the radio itself) sets the radio as selected. I was thinking that perhaps we'd use the `SlotProvider` and `useSlotProps` to do that?

src/components/index.tsx
CraigCav: Yes. We can follow the same template we use for `EzContent` (and the others).

src/components/index.tsx
CraigCav: Yes. It will likely be a similar format to the docs we use for EzContent (and the other slot components). However, we might need to adjust/modify to describe any other behaviors EzLabel has (such as its props and that it renders a `label` element).

I suspect we'll want to give some careful consideration as to whether the existing API for `EzLabel` is something we want exposed. Let's chat IRL about this.

src/components/index.tsx
danidewitt: added

danidewitt:
@CraigCav @pgrippi this is ready for final review now, then i'll rebase the commits and add a changeset

danidewitt:
@CraigCav @pgrippi this is ready for final review now

src/components/EzField/EzChoice.tsx
```diff
@@ -46,25 +47,43 @@ const box = theme.css({
...
-const Option = ({input, bordered, disabled, label}) => {
+const Option = ({input, bordered, disabled, label, labelFromChildren, contentFromChildren}) => {
   const id = useUniqueId();
+  const childLabelWithHtmlFor = React.cloneElement(labelFromChildren, {htmlFor: id});
   return (
     <Style ruleset={theme}>
       <span className={box({bordered, disabled})}>
         {React.cloneElement(input, {id})}
-        <label htmlFor={id}>{label}</label>
+        {(labelFromChildren && childLabelWithHtmlFor) || <label htmlFor={id}>{label}</label>}
```
CraigCav, line 63:
```suggestion
        {(childLabelWithHtmlFor) || <label htmlFor={id}>{label}</label>}
```

Presumably if `labelFromChildren` is `undefined`, `childLabelWithHtmlFor` should also be `undefined`? Looking at this a little further, if `labelFromChildren` is `undefined` (or `null`), I would have expected the `React.cloneElement` call to fail. 

What am I missing?

CraigCav: I think this code relies on `labelFromChildren` being `false` (rather than `undefined` or `null`), otherwise `React.cloneElement` would throw: 

image

Passing `false` for `labelFromChildren` doesn't seem like it should be valid (if we had type defs for the arguments instead of `any`, the compiler would be mad about this).

CraigCav: FWIW, if we update `EzLabel` to use [`Slot`](https://github.com/ezcater/recipe/blob/main/src/components/EzContent/Slot.tsx), you could skip using `React.cloneElement` and use [SlotProvider](https://github.com/ezcater/recipe/blob/main/src/utils/slots.ts#L21-L41) to wire up the `htmlFor` attribute:

```diff
const EzLabel = forwardRef<Ref, Props>(({as: element, ...initialProps}, ref) => {
  const {props} = styles(initialProps);

  return (
    <Style ruleset={theme}>
--      <Element {...props} ref={ref} />
++      <Slot element={element} slot="label" {...props} ref={ref} /> 
    </Style>
  );
});

```
We can also use the `SlotProvider` to wire up the `className` for the content (rather than wrapping with an additional `div`):

```diff
const Option = ({input, bordered, disabled, label, labelFromChildren, contentFromChildren}) => {
  const id = useUniqueId();
  return (
++  <SlotProvider slots={{label: {htmlFor: id}, content: {className: nestedContent()}}}>
      <Style ruleset={theme}>
        <span className={box({bordered, disabled})}>
          {React.cloneElement(input, {id})}
--        {(labelFromChildren && childLabelWithHtmlFor) || <label htmlFor={id}>{label}</label>}
++        {(labelFromChildren) || <Slot element="label" slot="label">{label}</Slot>}
        </span>
--      {contentFromChildren && <div className={nestedContent()}>{contentFromChildren}</div>}
++      {contentFromChildren}
      </Style>
++  </SlotProvider>
  );
};
```

src/components/EzField/EzChoice.tsx
```diff
@@ -46,25 +49,48 @@ const box = theme.css({
...
-const Option = ({input, bordered, disabled, label}) => {
+const Option = ({input, bordered, disabled, label, labelFromChildren, contentFromChildren}) => {
   const id = useUniqueId();
   return (
     <Style ruleset={theme}>
-      <span className={box({bordered, disabled})}>
-        {React.cloneElement(input, {id})}
-        <label htmlFor={id}>{label}</label>
-      </span>
+      <SlotProvider slots={{label: {htmlFor: id}, content: {className: nestedContent()}}}>
```
danidewitt, line 62: this is pretty nice @CraigCav

PR #658, opened 2021-06-14
test: avoid spec failures caused by scroll-snapping

CraigCav:
## What did we change?

test: avoid spec failures caused by scroll-snapping

- removed scroll snapping from visual regression test runs to avoid an issue where the browser would _sometimes_ trigger a scroll-snap during the initial paint of the carousel, causing the carousel to jump to the last page instead of showing the first page.

## Why are we doing this?

Our CI runs occasionally fail on the `Space for peek is reapplied across breakpoints` spec, showing that the carousel is rendered on the last page (items 9-10) instead of the second page (items 3-4).

## Screenshot(s) / Gif(s):

Failing snapshot:

image

## Checklist


danidewitt (approved):
i assume you ran this enough times to have run into a flakey case that didn't fail but used to?

PR #663, opened 2021-06-21
Use published version of stitches instead of ES5 build

CraigCav:
## What did we change?

- This branch moves Recipe onto using the published NPM build of `@stitches/core@0.1.0-canary.20` rather than our local copy (which is transpiled to ES5 for IE11 support).
- Added a GitHub action workflow step to allow deployment based on the presence of a PR label

## Why are we doing this?

- Cutting a version of Recipe for load testing to rule out whether using a transpiled version of stitches has introduced a significant overhead.

NOTE: during SSR, we needn't use a transpiled version of stitches (as Node is able to run modern JS), so if the ES5 version proves to be significantly slower, it's potentially possible to eliminate this overhead.

## Checklist


CraigCav:
@pgrippi Load tested this change for comparison against `main`. There was no significant difference to the load test run.

PR #665, opened 2021-06-22
adds recommended extensions in .vscode

ez-alexfrazer:
## What did we change?

- Added `.vscode/extensions.json`

## Why are we doing this?

Primarily for a more friendly developer experience. It will hint common extensions.

This is nice particularly for formatters; if the pipeline requires passing eslint, this will allow users to integrate it more quickly into their environments

There's also kind of the problem that there are lots of available packages for any task. This is a decent way to provide "best practices"

https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

## Checklist

- [] Create a changeset (`npm run changeset`) with a summary of the changes


.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["dbaeumer.vscode-eslint", "rvest.vs-code-prettier-eslint", "pmneo.tsimporter"]
```
pgrippi, line 2: I think we should switch `rvest.vs-code-prettier-eslint` for `esbenp.prettier-vscode`, it is no longer recommended to run prettier through eslint and it works just fine with recipe

pgrippi (dismissed):

.vscode/extensions.json
noranda: Agreed! Will make the change and rebase.

pgrippi (approved):

noranda (approved):

danidewitt (approved):
thank you!

PR #668, opened 2021-06-23
Use up-to-date build of stitches

CraigCav:
## What did we change?

- updated `@stitches/core` to latest (currently v0.2.2)
- updated local build of `@stitches/core` to be committed to `dist` to simplify Recipe's build (no longer have to build stitches in order to build Recipe).
- replaced `@ezcater/snitches` support latest version of `@stitches/core`
- changed Recipe's theme usage such that there is only a single theme, instead of one-per-component
  -  this was needed as stitches native style insertion assumes a single theme, and the sorting mechanism relies on this behavior
- fix: console warnings about `stroke-linecap` et al (using html attribute keys instead of react prop keys)
- Normalize snitches usage for EzBaseFontSizeCompatibility (to ensure it's styles would continue to be inserted correctly
- fix: ensure `EzProvider` is always a style provider (to make it easy for apps to have a top-level style provider)
- Update Recipe with of the stitches breaking API changes.
  - renamed `conditions` to `media` on theme and within component usage
  - added `@` prefix to responsive props (internally as not to change Recipe's public API)
  - Use new stitches API for disabling stitches style insertion (changed from `insertMethod` to changing the root to `null`)
  - Renamed hyphenated utilities
  - Removed changed/nested CSS function usage
- Simplified `EzPageHeader` styles using variants. While investigating non-deterministic style ordering, I noticed an opportunity to simplify how we're defining the page header styles.

## Why are we doing this?

Load testing of store picked up performance issues within Recipe v12 that are resolved when using an up-to-date version of stitches.

## Checklist


doc-site/src/components/Layout.js
```diff
@@ -126,12 +125,7 @@ const Layout = ({
               rel="stylesheet"
             />
           </Helmet>
-          <ThemeProvider
-            theme={{
-              ...themes.standard,
-              breakpoints: {medium: '768px', large: '1061px', xlarge: '1280px'},
-            }}
-          >
```
CraigCav, line 134: This hasn't been used in a while - Recipe doesn't use emotion's theme provider anymore, and the doc-site doesn't use responsive styles (it relies on Recipe layout instead).

doc-site/src/components/Playground/Preview.js
```diff
@@ -16,9 +16,9 @@ const codeIcon = (
     stroke="currentColor"
   >
     <path
-      stroke-linecap="round"
-      stroke-linejoin="round"
-      stroke-width="2"
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      strokeWidth="2"
```
CraigCav, line 21: These lint-fixes are unrelated to the stitches upgrade, but the warnings in the console were making it harder to debug.

danidewitt: thank you!

src/components/EzPageHeader/EzPageHeader.tsx
```diff
@@ -165,7 +155,7 @@ const EzPageHeader: React.FC<HeaderProps> = ({
           </MaxWidth>
         </div>
         {subheader && (
-          <div className={clsx(pageHeader(), subheaderWrapper())}>
+          <div className={clsx(pageHeader({subHeader: true}))}>
```
danidewitt, line 158: 👍 nice change, this pattern is so much nicer

PR #669, opened 2021-06-30
[DEVEX-1339] Update EzLabel

danidewitt:
## What did we change?
- Deprecated `EzLabelledItem`
- Improve `EzLabel` and expose it as a top-level component

## Why are we doing this?
- The main reason we prioritized this work was for the radio and checkbox fields slots work. See [description](https://ezcater.atlassian.net/jira/software/projects/DEVEX/boards/131?selectedIssue=DEVEX-1339). 

## Screenshot(s) / Gif(s):
Screen Shot 2021-07-07 at 10 24 03 AM

Screen Shot 2021-07-07 at 2 50 16 PM

## Checklist


src/components/EzField/EzField.tsx
```diff
@@ -51,11 +51,11 @@ const field = theme.css({
   variants: {
     labelSize: {
       small: {'& > * + *': {marginTop: '$50'}},
-      normal: {'& > * + *': {marginTop: '$100'}},
+      field: {'& > * + *': {marginTop: '$100'}},
```
CraigCav, line 54: As it stands, this would be a breaking change - anyone using `size: normal` would potentially break (or not be styled correctly) due to this.

I think this is actually find to stay as `normal`; `labelSize: field` doesn't give us any additional meaning as the component itself is already a field.

danidewitt: what do you think of [this fix](https://github.com/ezcater/recipe/pull/669/commits/b85c9c1ffae0ed37ad8f1bfa991083ebd9059676)? i was trying to get all the label size types to get along properly without making any breaking changes.

danidewitt:
@CraigCav i've reverted the margin to padding [change](https://github.com/ezcater/recipe/pull/669/commits/e9d3d0169b3e474eca5987b23467fb7c7382e79d) because one of the `EzLabelledItem` tests were failing (`Top Label`), and reverting just that change fixed it? 🤷‍♀️

src/components/EzField/EzField.tsx
CraigCav: I'm more thinking along the lines of:

```ts
// src/components/EzField/EzField.types.ts
labelSize?:  'normal' | 'small';
```

and:

```ts
// src/components/EzField/EzField.tsx
use={props.labelSize === 'small' ? 'small' : 'field'}
```

CraigCav: That being said, `small` and `field` really does feel like a mismatched pair of values. Do you have any thoughts on better alternatives?

Maybe something like `strong` and `subdued` (like EzTextStyle) would be a better fit?

CraigCav: Or perhaps `primary` and `secondary`? I didn't like the idea of this at first, but I like that it focuses on role, rather than style. Effectively "these fields are the main content, these other fields are ancillary" rather than "these are bold, these are grayed out".

danidewitt: a note if we come back to this: we settled on `primary` and `secondary` after discussions amongst ourselves and designers

src/components/EzLabel/EzLabel.tsx
```diff
@@ -4,19 +4,45 @@ import theme from './EzLabel.theme.config';
...
 const styles = theme.css({
   display: 'block',
-  lineHeight: '$label',
-  fontWeight: '$label',
-  padding: 0,
 
   variants: {
-    size: {
+    use: {
+      /*
```
danidewitt, line 10: it seems `EzLabel` should be the only thing that cares about the legacy style names, not the downstream components using it, like `EzField` and `EzLabelledItem`. it also seems like the easiest, least complex thing to do here is to just duplicate the legacy and new style names, to reduce complexity when using the `use` prop and also to reduce complexity when it comes to all the backwards compatibility with the type definitions.

if there's a nicer way to do this, that is still easy to read and understand, i'm open to it. obviously it's not ideal to repeat the `use` variants. maybe there's a nice, readable way to make `EzLabel` take all 4 possible props `normal`, `small`, `primary`, `secondary` and map them to the correct new stitches styles without duplicating the legacy styles?

src/components/EzField/EzField.tsx
```diff
@@ -52,6 +52,8 @@ const field = theme.css({
     labelSize: {
       small: {'& > * + *': {marginTop: '$50'}},
       normal: {'& > * + *': {marginTop: '$100'}},
+      primary: {'& > * + *': {marginTop: '$100'}},
```
danidewitt, line 55: i don't love this? 🤷‍♀️

src/components/EzField/EzField.tsx
```diff
@@ -174,7 +176,7 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
                 htmlFor={isChoiceElement ? undefined : id}
                 as={labelType}
                 error={showError}
-                size={props.labelSize}
+                use={props.labelSize || 'primary'}
```
danidewitt, line 179: i also don't love this, but it was complaining without the second condition: 

```
Panicking because nodes appear to be being changed every time we run queries. This would cause the site to recompile infinitely.
Check custom resolvers to see if they are unconditionally creating or mutating nodes on every query.
This may happen if they create nodes with a field that is different every time, such as a timestamp or unique id.
```

src/components/EzLabel/EzLabel.preview.md
```diff
@@ -0,0 +1,14 @@
+```jsx
```
danidewitt, line 1: remind me, how do i get this to generate a preview image?

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
+---
+name: EzLabel
+title: Labelling
```
danidewitt, line 3: `EzLabelledItem` title is also `Labelling` - do we want them to be the same?

src/components/EzLabelledItem/EzLabelledItem.md
```diff
@@ -6,6 +6,15 @@ path: '/components/ez-labelled-item'
 tags: ['label']
 ---
 
+<EzAlert
```
danidewitt, line 9: is there anything else we need to do to mark this as deprecated?

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+### With input controls
+
+Use `htmlFor={controlId}` to associate the label with the form control. This allows the label to increase the hit-surface to focus/activate the input, as well as allowing assistive technology to identify the label when the input is focused.
+
+```jsx
+<>
+  <EzLabel htmlFor="search-input" use="secondary">
+    Search
+  </EzLabel>
+  <EzSearchInput id="search-input" placeholder="Search customers" />
```
danidewitt, line 55: i don't love the extra margin above this search input, but it's not blocking

src/components/EzLabel/EzLabel.preview.md
danidewitt: Screen Shot 2021-07-07 at 10 27 18 AM

src/components/EzLabel/EzLabel.md
danidewitt: Screen Shot 2021-07-07 at 10 27 18 AM

src/components/EzLabelledItem/EzLabelledItem.md
danidewitt: Screen Shot 2021-06-30 at 1 34 57 PM

src/components/EzLabel/EzLabel.md
danidewitt: Screen Shot 2021-07-07 at 10 57 49 AM

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+
+Labelled items should
+
+#### Do
+
+- Use prop `use="primary"` to label important content, like form fields.
+- Use prop `use="secondary"` to label supporting content, like a search field. Secondary labels are used to avoid drawing attention away from primary content.
+- Use `EzLabel` without a `use` prop to label items without styling, in other words, with default body styling.
+- Use `htmlFor={controlId}` to associate the label with the form control. This allows the label to increase the hit-surface to focus/activate the input, as well as allowing assistive technology to identify the label when the input is focused.
+
```
danidewitt, line 37: do we want to document the position options? do we want to document how to use `as='label'` and `as='div'`?

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+## Content guidelines
+
+Labelled items should
+
+#### Do
+
+- Use prop `use="primary"` to label important content, like form fields.
+- Use prop `use="secondary"` to label supporting content, like a search field. Secondary labels are used to avoid drawing attention away from primary content.
+- Use `EzLabel` without a `use` prop to label items without styling, in other words, with default body styling.
+- Use `htmlFor={controlId}` to associate the label with the form control. This allows the label to increase the hit-surface to focus/activate the input, as well as allowing assistive technology to identify the label when the input is focused.
```
CraigCav, line 36:
```suggestion
- Use `htmlFor={controlId}` to associate the label with the form control. This allows the label to increase the hit-surface to focus/activate the input, as well as allowing assistive technology to identify the label when the input is focused. Alternatively, an `<input>` element can be nested directly inside the label to automatically associate the label with the input element.
```

src/components/EzLabel/EzLabel.md
CraigCav: Yes, but probably not under the `do` section. The `do` section is intended as content guidelines - basically a means to communicate the _why_ behind the API. 

Perhaps if we change the `### Side label` example to `### Label positioning`, it'll give us a place we can talk about the other available options.

src/components/EzLabel/EzLabel.md
CraigCav: I _think_ the extra margin here is coming from the card wrapping the example. Does the margin go away if you change the `<>` fragment wrapper for a `<div>`?

I would imagine the margin would also disappear if you did this::
```jsx
<EzLabel use="secondary">
  Search
  <EzSearchInput placeholder="Search customers" />
</EzLabel>
```

...but that's only if you want to demonstrate the other way you can associate a label with a form element (which we should probably also do).

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Primary label
+
+A label for important content.
+
+```jsx
+<>
+  <EzLabel use="primary">Primary Label</EzLabel>
+  Some text
```
CraigCav, line 66: I think we should be careful to demonstrate labels in ways we'd expect them to be used in downstream applications. A label without an associated input is problematic from an accessibility standpoint (given that it renders a `<label>` element).

Perhaps we could use an input here instead:
```suggestion
  <EzLabel use="primary" htmlFor="id="primary-input">Primary Label</EzLabel>
  <input id="primary-input" placeholder="Some text" />
```

or demonstrate the `as` prop - we'll have to be careful to specifically call out _why_ the `as` prop is needed:
```suggestion
  <EzLabel use="primary" as="span">Primary Label</EzLabel>
  Some text
```

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+```jsx
+<>
+  <EzLabel position="side" use="primary">
+    Side Primary Label
+  </EzLabel>
+  Some text
+</>
+```
+
+### Bottom label
```
CraigCav, line 94: Thought's on killing the `### Bottom label` example and combining it with `### Side label` as `### Label positioning`?

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+
+A label that appears below its content.
+
+```jsx
+<>
+  Some text
+  <EzLabel use="primary">Bottom Primary Label</EzLabel>
```
CraigCav, line 101: Is this missing `position="bottom"`?

src/components/EzLabel/EzLabel.preview.md
CraigCav: I created a node script to do that in [this PR](https://github.com/ezcater/recipe/pull/494/files). You have to create a markdown file with an example that you want to create a snapshot of (similar to our snapshot test setup), then run the CLI.

That being said, I think we can repurpose the `EzLabelledItem` image here (i.e. rename the image we use for EzLabelledItem from the doc-site to `EzLabel.png`) and remove entry for `EzLabelledItem`.

src/components/EzLabel/EzLabel.tsx
```diff
@@ -4,34 +4,36 @@ import theme from './EzLabel.theme.config';
...
+      secondary: {
+        color: '$label-small',
+        fontSize: '$label-small',
+        lineHeight: '$label',
+        fontWeight: '$label',
+        padding: 0,
       },
     },
     position: {
       hidden: {'sr-only': 'true'},
```
CraigCav, line 26: We should probably add documentation for this use case. We'll want to call out that it's used to programmatically associate an input element with a label for accessibility purposes.

src/components/EzLabel/EzLabel.tsx
```diff
@@ -4,34 +4,36 @@ import theme from './EzLabel.theme.config';
...
       top: {marginBottom: '$label-my'},
       bottom: {marginTop: '$label-my'},
-      left: {marginRight: '$label-mx', display: 'inline-block'},
-      right: {marginLeft: '$label-mx', display: 'inline-block'},
+      side: {marginRight: '$label-mx', display: 'inline-block'},
     },
     error: {true: {}},
   },
   compoundVariants: [{error: 'true', css: {color: '$negative'}}],
```
CraigCav, line 33: Should we document this too?

src/components/EzLabel/EzLabel.tsx
```diff
@@ -68,14 +70,14 @@ const EzLabel = forwardRef<Ref, Props>(({as: Element, ...initialProps}, ref) =>
...
- * @property {'hidden' | 'top' | 'bottom' | 'left' | 'right'} position - label renders on top by default
- * @property {'normal' | 'small'} position - label renders with normal size by default
+ * @property {'label' | 'legend' | 'div'} as - label renders as a label element by default
+ * @property {'top' | 'bottom' | 'side' | 'hidden'} position - label renders on top by default
  */
 EzLabel.defaultProps = {
-  as: 'div',
   position: null,
-  size: 'normal',
+  as: 'label',
```
CraigCav, line 78: We don't provide any documentation for the `as` prop yet, but it's important to call out for accessibility reasons (rendering a label without an associated input causes accessibility issues).

tl;dr is that we use the `as` prop when we want to render an element that _looks_ like a label, without the label's baked in accessibility behaviors.

src/components/EzLabelledItem/EzLabelledItem.md
CraigCav: We're a bit inconsistent about this today I think, but I think it's a good idea to mark the component with doc-comments (like we do for [`EzContentGroup`](https://github.com/ezcater/recipe/blob/3f242b1a9f71feefddb4d4fd1e0e00d9150aa28f/src/components/EzPageContent/EzContentGroup.tsx#L9-L11)). 

This way IDEs (and linters) will let developers know that the component is deprecated:

image

src/components/EzSegmentedControl/__tests__/EzSegmentedControl.test.tsx
```diff
@@ -62,7 +62,7 @@ describe('EzSegmentedControl', () => {
   const active = true;
   const className = 'string';
   const label = (
-    <Label id="1" position="top">
+    <Label id="1" position="top" use="primary" as="div">
```
CraigCav, line 65: I might be wrong, but looking at the intent of this test (typechecking EzSegmentedControl), I would image the idea here is to test that the EzSegmentedControl component can accept arbitrary `ReactNode` content in place of a text label. 

In which case, can probably simplify this text to passing a `div` here instead of using the `EzLabel` component.

If that's NOT what we're trying to test, then this label should probably just be a string.

src/components/EzToggle/EzToggle.tsx
```diff
@@ -119,13 +119,15 @@ const EzToggle = forwardRef<HTMLInputElement, Props>(({id, label, status, ...pro
...
           ref={ref}
         >
-          <input {...inputProps} id={idOrDefault} ref={inputRef} type="checkbox" />
-          <div className={track()} />
-          {label && (
-            <EzLabel htmlFor={idOrDefault} position="right" as="label">
-              {label}
-            </EzLabel>
-          )}
+          <EzLayout layout="basic">
```
CraigCav, line 122: I think this can be simplified a little, as the EzToggle component is already using `EzLayout`.

If you move the `label && (...)` code down outside the div, I believe the layout will just work without adding an extra wrapper. You may then have to target the `EzLayout` item with styles to ensure that the disabled state still reduces the text opacity for the label of the disabled element.

CraigCav (approved):

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+## Best Practices
+
+Labelled items should:
+
+- Be used to label a static piece of text
+- Be used with the `htmlFor` attribute to label form controls.
+
+Labelled items should not:
+
+- Be used to label more than one child (unless being used with a single child that is a collection of multiple items, such as a Segmented Control)
```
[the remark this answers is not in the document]
CraigCav: Ah gotcha, thanks for the context.

I think `should only be associated with a single form input` is a good way to phrase this. We can definitely remove the Segmented Control text - it's more confusing than it is helpful.

src/components/EzField/EzField.tsx
```diff
@@ -174,7 +174,13 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
...
                 as={labelType}
                 error={showError}
-                size={props.labelSize}
+                use={
+                  props.labelSize === 'normal'
+                    ? 'primary'
+                    : props.labelSize === 'small'
+                    ? 'secondary'
+                    : props.labelSize || 'primary'
+                }
```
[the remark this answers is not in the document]
danidewitt: yeah, good catch. i was originally just trying to map it like this `const sizes = { small: 'secondary', normal: 'primary', }` with `use={sizes[size] || size}` but that had interesting type issues...

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,103 @@
...
+- Be used to label components that already contain labels, such as EzField. Use the label property of the component directly when possible e.g. `EzField[label={label}]`
```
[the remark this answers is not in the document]
danidewitt: haha thanks for the credit - i didn't write that - it was from `EzLabelledItem`

src/components/EzLabel/EzLabel.md
danidewitt: does it need it? it's already below it?

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+    Bottom Primary Label
+  </EzLabel>
+</div>
+```
+
+Hidden labels, with `position='hidden'`, are used to programmatically associate an input element with a label for accessibility purposes.
+
+```jsx
+<div>
+  <EzLabel position="hidden" htmlFor="hidden-content">
```
CraigCav, line 121: Nitpick: it's not the content that is hidden, it's the label, so this `Id` value is misleading.

Specifically we're trying to show how folks can use `EzLabel` to give assistive technology extra cues about the purpose of a given input element when the label would otherwise be visually hidden. Typically you'll find labels are omitted from inputs when the context is otherwise clear from the page structure/layout - something that may not be so clear for visually impaired users.

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+### Labels with error state
+
+When provided the `error` prop, the label will receive error styling.
+
+```jsx
+<div>
+  <EzLabel use="primary" error={true}>
+    Label
+  </EzLabel>
+  Some text
```
CraigCav, line 137: Please use an input here too.

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+</div>
+```
+
+### Label accessibility
+
+Labels render `<label>` elements by default, and with the `as='label'` prop, which provide accessibility behaviors. To render an element that looks like a label without accessibility behaviors, provide another element to the `EzLabel` prop `as`, examples: `as='div'` or `as='span'`.
```
CraigCav, line 143:
```suggestion
Labels render `<label>` elements by default, allowing form elements to be programmatically associated with an identifying text caption and allows browsers to provide additional user experience benefits such as using the label to provide a larger hit-target for inputs.

To render an element that looks like a label without rendering a html `<label>` element, the `as` prop can be provided with another element type, such as: `as='div'` or `as='span'`.

This may be useful if providing a top-level label for a group of inputs where each individual input may already have a label, such as a radio button group.

NOTE: the `htmlFor` prop should be omitted when using the `as` prop. In order to retain programmatic association of the label with the content, additional `aria-` attributes such as `aria-labelledby` may be necessary.
```

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+```jsx
+<div>
+  <EzLabel htmlFor="text-container">Secondary Label</EzLabel>
+  <div id="text-container">Some text</div>
+</div>
+```
+
+### Label positioning
+
+A label that appears next to its content.
```
CraigCav, line 95:
```suggestion
Top labels are the default position for `EzLabel` and are recommended because they work better with long copy, localization, and responsive layouts. Side labels can be useful when vertical space is limited.

The label's overall position relative to the element it is labeling may be changed using the `position` prop.
```

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+```jsx
+<div>
+  <EzLabel position="side" use="primary">
+    Side Primary Label
+  </EzLabel>
+  Some text
+</div>
+```
+
+A label that appears below its content.
```
CraigCav, line 106: Let's skip the `position=bottom` example.

src/components/EzLabel/EzLabel.md
```diff
@@ -0,0 +1,152 @@
...
+### Label positioning
+
+A label that appears next to its content.
+
+```jsx
+<div>
+  <EzLabel position="side" use="primary">
+    Side Primary Label
+  </EzLabel>
+  Some text
```
CraigCav, line 102: Let's use an input here.

src/components/EzLabelledItem/EzLabelledItem.tsx
```diff
@@ -13,6 +13,9 @@ type ItemProps = {
   htmlFor?: string;
 };
 
+/**
+ * @deprecated Please use EzPageSection or EzLayout instead.
```
CraigCav, line 17:
```suggestion
 * @deprecated Please use EzLabel instead.
```

danidewitt:
@CraigCav is there anything left to do here, besides rebase and merge?

PR #670, opened 2021-07-12
Compatibility with apps using "strictNullChecks": true

CraigCav:
Demo: https://codesandbox.io/s/recipe-with-ts3-and-strictnullchecks-i478s

## What did we change?

- Switched from exposing stitches `VariantsCall` type definition for responsive props within Recipe, to using a custom `VariantProps` type derived from the stitches style definitions.

## Why are we doing this?

- Stitches relies on TypeScript features only available in TS 4+ (template literal types). Exposing these types in our public API causes issues for Recipe apps that are on older versions of Typescript. 
- Recipe was relying on it's internal configuration using `"strictNullChecks": false` when generating type definitions for responsive props. Recipe used a `keyof` call upon a nullable type, which returns `unknown` when called on a type that may be `undefined` when running in strict mode. In non-strict mode, `keyof` would return the possible keys of the type, ignoring the possibility that the type may be `undefined`. The new `VariantProps` type definition avoids calling `keyof` on nullable types, and is therefore safe for use within strict mode.

## Checklist


src/components/EzHeading/EzHeading.tsx
```diff
@@ -11,27 +13,27 @@ const headingCss = theme.css({
 
   variants: {
     size: {
-      1: {
+      '1': {
```
CraigCav, line 16: Using `number` here instead of `string` would be correct if `EzHeading` expected the `size` prop to be a number, however we've [historically always expected a string value here](https://github.com/ezcater/recipe/blob/3f242b1a9f71feefddb4d4fd1e0e00d9150aa28f/src/components/EzHeading/__tests__/EzHeading.test.tsx#L58-L66) (and changing that would break compatibility). I've had to disable the `quote-props` lint rule for this file in order to stop eslint recommending that we break this code 😄 

FWIW, we didn't bump into this when exposing Stitches `VariantCall` directly, since stitches internally maps prop types to a pretty [fuzzy type definition](https://github.com/modulz/stitches/blob/1fa120b8c29789d5ef464958bdb5e9e493b16fe8/packages/core/types/core.d.ts#L481) where boolean values can be provided as `true | false | 'true' | 'false'` and number values can be provided with a string equivalent value i.e. `1 | 2 | 3 | '1' | '2' | '3'`. Note also, stitches uses Typescript [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to achieve this fuzzy type definition, but this feature relies on TS 4.1 and higher.

danidewitt (approved):

PR #680, opened 2021-07-23
fix: ignored id prop on EzField

### LINKED ISSUE #433: Bug: EzField ignored ID props
## Describe the bug

Reported by the store team: https://ezcater.slack.com/archives/CBC9DMGP5/p1602081102027700

> I am passing an actual id and name, but ezfield is ignoring it (and the onchange handler formik binds won’t work because it requires name match the actual form field)

Update:

[It was noted that formik was being used incorrectly](https://github.com/ezcater/recipe/issues/434). Once corrected, only the `id` prop is being overridden by EzField.

## Steps To Reproduce

1.
2.

Link to code example:

CraigCav:
## What did we change?

- fix: ignored id prop on EzField

## Why are we doing this?

- Closes #433

## Checklist


danidewitt (approved):
🚀

PR #681, opened 2021-07-23
Fix textvalue warning

CraigCav:
## What did we change?

- fix: missing `textValue` warning from Multiple choice inputs
- fix: disabled choices via slots API
- refactor: simplify logic for mapping between options and slots

## Why are we doing this?

- Fixes console warnings when running test suite.

## Checklist


src/components/EzField/EzChoice.tsx
```diff
@@ -49,26 +50,30 @@ const box = theme.css({
   ],
 });
 
+const label = theme.css({
+  maxWidth: 'calc($full - $300)',
```
danidewitt, line 54: 👍

src/components/EzField/EzChoice.tsx
```diff
@@ -49,26 +50,30 @@ const box = theme.css({
...
+        slots={{label: {htmlFor: id, className: label()}, content: {className: nestedContent()}}}
+      >
         <span className={box({bordered, disabled})}>
           {React.cloneElement(input, {id})}
-          {labelFromChildren || (
-            <Slot element="label" slot="label">
-              {label}
-            </Slot>
-          )}
+          {rendered}
```
danidewitt, line 75: 👍

src/components/EzField/EzChoice.tsx
```diff
@@ -114,17 +119,15 @@ export default props => {
...
-          const label = choice.label;
-          const labelFromChildren = optionHasChildren
-            ? optionChildren.find(child => child.type?.displayName === 'EzLabel')
-            : undefined;
-          const contentFromChildren = optionHasChildren
-            ? optionChildren.find(child => child.type?.displayName === 'EzContent')
-            : undefined;
-          const value = choice.value || choice.key;
-          const disabled = props.disabled || choice.disabled;
+          const rendered = choice.label ? (
```
danidewitt, line 122: 👍 this refactoring is nice

src/components/EzField/EzField.md
```diff
@@ -212,7 +212,7 @@ Alternatively, instead of using the `options` prop, EzField also accepts `<EzIte
 
 Allows for additional content to be presented with each choice of a multiple choice selection.
 
-EzField accepts `<EzItem>` elements as children, each with a `key` prop to denote the item value. Note, that `EzItem` must be a **direct descendant** of the field, and not wrapped in additional DOM nodes or React components.
+EzField accepts `<EzItem>` elements as children, each with a `textValue` prop to denote the item value. Note, that `EzItem` must be a **direct descendant** of the field, and not wrapped in additional DOM nodes or React components.
```
danidewitt, line 215: thank you!

src/components/EzField/__tests__/EzField.radio.test.md
```diff
@@ -54,3 +54,19 @@
   />
 </Media>
 ```
+
+### Multiple choice options can be disabled via slots API
```
danidewitt, line 58: 👍

src/components/theme.config.ts
```diff
@@ -122,6 +122,7 @@ const stitches = createCss({
       50: '4px',
       150: '12px',
       200: '16px',
+      300: '24px',
```
danidewitt, line 125: um... was this not here before?

CraigCav: It's available as a `space` value, but not a `size` value. Stitches allows space and size values to differ, but realistically I think it makes sense for us to keep them consistent. We should probably have a think about whether it makes sense to collapse these values into a single variable somehow to minimize the size of the css output we generate.

PR #684, opened 2021-07-27
New Component: EzWell

CraigCav:
## What did we change?

- Add new component EzWell 
  - A Well is a content container used to visually group content on a page. Typically a Well displays non-editable content separate from other content on the screen.
- Moved doc-site sprites from `playground/preview` into `layout` so they're only inserted once per page
- Updated doc-site to use Well instead of custom EzCard
- Updated EzStatus example to use EzLayout for spacing (it previously  implicitly relied on EzCard styles to space out the paragraphs)
- Disabled `react/require-default-props`- forcing a "default" for optional values... kinda defeats the point of optional values..

## Why are we doing this?

We use this pattern in a handful of places today, so this work is a consolidation of an existing pattern to promote consistency.

## Screenshot(s) / Gif(s):

image

image

The quote is from H.G. **Well**s. Geddit? :smiley_cat: :dizzy_face:

## Checklist


doc-site/src/components/Layout.js
```diff
@@ -6,7 +6,6 @@ import {StaticQuery, graphql, withPrefix} from 'gatsby';
 import Link from 'gatsby-link';
 import naturalSort from 'natural-sort';
 import {
-  themes,
```
danidewitt, line 9: for another PR - there must something we can configure to more obviously warn us of unused imports

src/components/EzWell/EzWell.tsx
```diff
@@ -0,0 +1,61 @@
+import React, {forwardRef, Ref, ReactNode, HTMLAttributes} from 'react';
+import Style from '@ezcater/snitches';
+import theme from './EzWell.theme.config';
+import {filterDOMProps} from '../EzField/filterDOMProps';
+
+const styles = theme.css({
```
danidewitt, line 6: do we want this to be named `styles`? at one point i thought we were trying to be more descriptive for the component theme names for debugging purposes? also we also talked about making separate styles files, like `EzWell.styles.tsx`. it obviously works as is. plus, we haven't been super consistent about either of these things in other places.

probably worth discussing/deciding on something consistent, at least, so we can make it more consistent in other places when we do future work in those files.

src/components/EzWell/EzWell.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const {children, role, ...otherProps} = props;
+
+  if (!role && (props['aria-label'] || props['aria-labelledby']))
+    console.warn('A labelled Well must have a role.');
+
+  return (
+    <Style ruleset={theme}>
+      <div
+        {...filterDOMProps(otherProps, {
+          propNames: new Set(['id', 'className']),
```
danidewitt, line 43: is `Set` an issue for IE11? or do we take care of that with snitches?

doc-site/src/components/Layout.js
CraigCav: There is already; lint _warns_ us about unused imports (but it doesn't blow up the build). Ideally though we'd configure our local environments to auto-fix this type of issue.

src/components/EzWell/EzWell.tsx
CraigCav: My take on this is that we _do_ want to be as-descriptive-as-possible with names, often collecting styles into named groups that represent the concept (like "arrow" styles on a callout). This is particularly true of conditional styles (or variant styles), as we want to name the condition or variant (like "bordered" or "accented" etc).

In this case though, the styles represent the complete set of styles for the whole component. 

We could call this variable `wellStyles` but given this file is already specific to `Well`, the prefix is redundant and a potential obstacle to future renaming.

PR #687, opened 2021-08-03
[DF-29] Add edit (from github) button to doc-site

danidewitt:
## What did we change?
- Add an "Edit" button to edit docs from github

## Why are we doing this?
- Encourage users to suggest edits to the doc-site

## Screenshot(s) / Gif(s):
Screen Shot 2021-08-03 at 4 58 54 PM

Screen Shot 2021-08-03 at 4 59 09 PM

Screen Shot 2021-08-03 at 5 00 34 PM

## Checklist

### DIFF
--- a/doc-site/src/components/EditDocsButton.js
+++ b/doc-site/src/components/EditDocsButton.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import {EzTooltip, EzLink} from '@ezcater/recipe';
+
+const EditDocsButton = ({path}) => {
+  if (!path || !path.fileAbsolutePath) {
+    return null;
+  }
+
+  const gitPath = path.fileAbsolutePath.split('/recipe/')[1];
+
+  const gitLink = `https://github.com/ezcater/recipe/edit/main/${gitPath}`;
+
+  return (
+    <EzTooltip message="Edit docs on GitHub">
+      <EzLink>
+        <a href={gitLink} target="_blank">
+          <svg
+            height="16"
+            fill="#8b99a6"
+            viewBox="0 0 16 16"
+            width="16"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <use xlinkHref="#editPencil" />
+          </svg>
+        </a>
+      </EzLink>
+    </EzTooltip>
+  );
+};
+
+export default EditDocsButton;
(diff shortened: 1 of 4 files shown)


doc-site/src/components/EditDocsButton.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import {EzTooltip, EzLink} from '@ezcater/recipe';
+
+const EditDocsButton = ({path}) => {
+  if (!path || !path.fileAbsolutePath) {
+    return null;
+  }
+
+  const gitPath = path.fileAbsolutePath.split('/recipe/')[1];
```
danidewitt, line 9: we'll have to decide if we want this on just the `/components/` pages, or on all pages like `/cookbook.`, `/support/`, etc.

doc-site/src/components/Layout.js
```diff
@@ -170,6 +175,7 @@ const Layout = ({
                           value={searchTerms}
                           onChange={e => setSearchTerms(e.target.value.toLowerCase())}
                         />
+                        <EditDocsButton path={absolutePath} />
```
danidewitt, line 178: we'll have to decide where we want this to live. initial thoughts are putting in the header "actions" section, either to the right or left of the search bar:

Screen Shot 2021-08-03 at 4 58 54 PM
Screen Shot 2021-08-03 at 4 59 09 PM

however, we may want it to be a little quieter, maybe in a different location? thoughts?

CraigCav:
> we may want it to be a little quieter, maybe in a different location? thoughts?

What do you think of something like this? 

image

I've used a pencil icon instead of the edit text and given it the deemphasis color, and given each card a top level "Overview" title (since cards require a title in order to also have actions).

We could do this by adding something like this to the doc-site layout:

```jsx

const actions = {
  title: 'Overview',
  actions: `/*anchor with svg image here */`
};

<EzCardSection key={i} {...(index === 0 ? actions : {})}>{section}</EzCardSection>

```

doc-site/src/components/EditDocsButton.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const gitLink = `https://github.com/ezcater/recipe/blob/main/${gitPath}`;
```
CraigCav, line 11: Thoughts on making this the edit page directly?

```suggestion
  const gitLink = `https://github.com/ezcater/recipe/edit/main/${gitPath}`;
```


| Blob   |      Edit      | 
|----------|:-------------:|
| image |  image |

danidewitt:
@CraigCav 

Screen Shot 2021-08-13 at 10 42 10 AM
Screen Shot 2021-08-13 at 10 42 26 AM

CraigCav (approved):

Issue #196, opened 2019-08-08
Add display names to Recipe components to assist with debugging


CraigCav:
The following components don't render with their corresponding component name when inspected using React dev tools:

- EzBanner
- EzCheckbox
- EzField (renders the name `Field` not `EzField`)
- EzFormLayout
- EzLayout
- EzLink
- EzSearchInput
- EzSuperRadioButtons
- EzTextStyle
- EzToggle

Screenshot of EzToggle displaying `Unknown` in dev tools:

<img width="257" alt="Screen Shot 2019-08-08 at 8 54 03 AM" src="https://user-images.githubusercontent.com/109814/62705245-037c2000-b9bb-11e9-84c5-ea0505511f96.png">

Related links:
- https://reactjs.org/docs/react-component.html#displayname
- https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

PR #688, opened 2021-08-04
[DF-63] Add display names to components

danidewitt:
## What did we change?
- Added display names to `Ez-` components

## Why are we doing this?
- Assist with debugging, closes https://github.com/ezcater/recipe/issues/196

## Checklist

### DIFF
--- a/src/components/EzCard/EzCardHeading.tsx
+++ b/src/components/EzCard/EzCardHeading.tsx
@@ -39,6 +39,6 @@ const EzCardHeading: React.FC<HeadingProps> = ({title, subtitle, actions}) => {
   return <div className={box()}>{actions ? layout : heading}</div>;
 };
 
-EzCardHeading.displayName = 'EzCard';
+EzCardHeading.displayName = 'EzCardHeading';
 
 export default EzCardHeading;
--- a/.changeset/tidy-spies-cover.md
+++ b/.changeset/tidy-spies-cover.md
@@ -1,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+workflow: export default for anonymous functions to improve debugging
--- a/src/components/EzField/EzChoice.tsx
+++ b/src/components/EzField/EzChoice.tsx
@@ -79,7 +79,7 @@ const Option = ({input, bordered, disabled, rendered}) => {
   );
 };
 
-export default props => {
+const EzChoice = props => {
   const {
     type,
     name: fieldName,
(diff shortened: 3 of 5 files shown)


danidewitt:
@CraigCav I think the only one I didn't add a display name to was `EzItem` - bc [prop types](https://github.com/ezcater/recipe/blob/e8adbf10c77dab394e5c597bce747d7ce20a4410/src/components/EzContent/EzItem.tsx#L4).

CraigCav:
@danidewitt The main issue with https://github.com/ezcater/recipe/issues/196, is with default exports with no name:

```jsx
// 🛑  - default export is an anonymous function, so React can't infer a name
export default () => {
  return <button />;
}
```

Any of the following patterns will show up named in DevTools and warnings, without needing an explicit `displayName` added:

```jsx
// ✅ - Named export (function declaration)
export function Button() {
  return <button />;
}

// ✅ - Named export (arrow function)
export const Button = () =>  {
  return <button />;
};

// ✅ - Default export  (function declaration)
export default function Button() {
  return <button />;
}

// ✅ - Default export (arrow function - dev tools picks up the function's `.name`, which is inferred from the variable name)
const Button = () =>  {
  return <button />;
};

export default Button;
```

I **_think_** that in the majority of cases, we follow the last example (Default export of a named arrow function), but there may be some components that don't follow that pattern and slip through the cracks. We should be able to find those cases by searching for `export default () =>`. I'd prefer to avoid blanket adding `displayName` across the board, as we're then needlessly shipping extra bytes.

So it's really just `export default () =>` we have to worry about. Those:

- show up as Unknown in React dev tools
- show up as anonymous in stack traces
- won't be checked by react's linting rules (i.e. checking for `use{Hookname}` etc)
- won't work with certain react features, like Fast Refresh.

src/components/EzCard/EzCardHeading.tsx
```diff
@@ -39,6 +39,6 @@ const EzCardHeading: React.FC<HeadingProps> = ({title, subtitle, actions}) => {
   return <div className={box()}>{actions ? layout : heading}</div>;
 };
 
-EzCardHeading.displayName = 'EzCard';
+EzCardHeading.displayName = 'EzCardHeading';
```
danidewitt, line 42: This seemed like the wrong display name, so I fixed it.

danidewitt:
@CraigCav i think i've narrowed it down to the right components. let me know what you think?

CraigCav (approved):

CraigCav:
Let's create a separate card for adding a lint rule to catch these going forwards.

Issue #645, opened 2021-06-03
Support sorting columns with custom cell rendering


jbelang:
I have been looking into replacing ezManage's use of Semantic table with EzTable, but one thing stands in the way: "Sorting columns with custom cell rendering is currently not supported." We have all the different pieces needed for these tables - sorting, custom cells, pagination, actions in the header - but they don't all work together.

Ideally, EzTable would support sorting columns with custom cells so that we could convert:
- Customers table
- EzD preferences table
- PaymentDetails table
- Payments table
- Performance table

This will help us toward our goal of removing Semantic from ezManage and switching entirely to Recipe.

CraigCav:
[...]

PR #689, opened 2021-08-04
[DF-65] Update docs for eztable to sort columns with custom cell rendering

danidewitt:
## What did we change?
- Updated docs for eztable to sort columns with custom cell rendering

## Why are we doing this?
- We can support sorting cells with custom rendering, closes https://github.com/ezcater/recipe/issues/645

## Screenshot(s) / Gif(s):
Screen Shot 2021-08-04 at 12 30 21 PM

## Checklist


src/components/EzTable/EzTable.md
```diff
@@ -316,7 +316,9 @@ The following properties are required when using selection:
 
 Use column sorting to help users find items quicker in larger data sets. Column sorting is switched off for all columns by default. Table data can only be sorted by a single column at a time.
 
-To turn on column sorting, set the `sortable` flag for each column where sorting is supported. Sorting columns with custom cell rendering is currently not supported.
+To turn on column sorting, set the `sortable` flag for each column where sorting is supported.
+
+To sort cells with custom rendering, provide a `key` to each column. The cell will be sorted by the `key`, instead of the custom React element or component provided to `component`.
```
CraigCav, line 321: > To sort cells with custom rendering, provide a `key` to each column. The cell will be sorted by the `key`, instead of the custom React element or component provided to `component`.

I don't think we can say the above line; The table component isn't responsible for sorting the data, it only tells the consuming application that an updated sort on the data is needed and which column and order to sort upon. It's the `onSortClick` implementation that decides which column to sort by; it can use the `column` metadata to make this decision.

Perhaps we could say something like:

```diff
--To turn on column sorting, set the `sortable` flag for each column where sorting is supported. Sorting columns with custom cell rendering is currently not supported.
++To turn on column sorting, set the `sortable` flag for each column where sorting is supported. 

-- The client code must also provide an `onSortClick` function as a prop to `EzTable`. When the column header for a sortable column is clicked, `EzTable` will notify the client that sorting is requested by calling the provided `onSortClick` function. The function is called with the click event as the first argument, and an object with the properties `column` and `direction` as the second argument. `column` is the object representing the column being sorted and `direction` is a string whose value represents the direction the sort should use, either `asc` or `desc`. When sorting an unsorted column, the initial sort direction is `asc`.
++When a column header is pressed, the `onSortClick` callback will be called and provided with `column` metadata and the `direction` in which to sort the data, either `asc` or `desc` for ascending or descending respectively. When sorting a column for the first time, the initial sort direction will be ascending.

When providing pre-sorted data to EzTable, the relevant column should have a `defaultSort` value set to either `asc` or `desc` to indicate the sort order applied to the data.

++The following example performs client-side sorting, using the `key` metadata of a column to indicate which column to sort by.
```

danidewitt: @CraigCav updated

CraigCav: @danidewitt I think we need to remove the `To sort cells with custom rendering...` line entirely (this is what I was trying to say above).

Recipe doesn't own the sorting logic, so we can't claim to be sorting by the `key`. Recipe provides column metadata to the downstream app, and the app decides how to sort. In our doc-site example we're sorting by key, but this behavior is implementation specific, and sorting might be implemented differently within apps.

CraigCav (approved):

Issue #683, opened 2021-07-26
Enhancement: add `onClick` handlers to `EzCarousel`


morgannegagne:
The Store team has a had a few instances where we want to track a user's interaction with `EzCarousel`, specifically looking to track whether a user scrolls through the options. 

I'm not sure of the best naming, but it would be nice to have hooks into `onClickRight` and `onClickLeft` to see if a user clicks the right/left buttons to view different options.

CraigCav:
Perhaps something like `onPageChange` might be a good option; the user can scroll through the options with other interactions besides a button press, like a two-finger swipe, or by tabbing through the carousel content (as might be the case with clickable cards).

I guess the implementation may depend on the level of granularity you need; are you interested in the interaction itself (i.e. distinguishing between clicks, swipes and tabs), or just whether the user is viewing the options beyond the first page? I suspect maybe the latter?

morgannegagne:
Just whether the user is viewing options beyond the first page, which I hope simplifies things a bit. Thanks @CraigCav!

PR #690, opened 2021-08-05
[DF-66] Page change handlers for EzCarousel

danidewitt:
## What did we change?
- Added `onPagePrevious` and `onPageNext` handler props to `EzCarousel`

## Why are we doing this?
- Allow actions to fire when user interacts with `EzCarousel`, closes https://github.com/ezcater/recipe/issues/683

## Screenshot(s) / Gif(s):
Screen Shot 2021-08-05 at 4 32 12 PM

Screen Shot 2021-08-05 at 4 32 53 PM

## Checklist


morgannegagne (approved):
This is great! Thank you! 🎉

danidewitt:
It's complaining about `New snapshot was not written. The update flag must be explicitly passed to write a new snapshot.` - @CraigCav do we need a new snapshot?

CraigCav (commented):
Thanks for tackling this issue.

I think with a couple of changes we'll be able to land this for the store team. 

One challenge of triggering `onChange` callbacks within `useEffect` (or as the result of a `useEffect` call), is that we can end up with "changes" firing without user interaction, like during component initialization. Instead, we should try to always call `onChange` callbacks as a direct result of user interaction - typically within an event handler like `onClick` or `onScroll`.

We might also want to consider simplifying the API down to just `onPageChange` rather than necessarily `onPagePrevious` and `onPageNext` - the use case only calls to identify changes to the page, so by splitting this into two, we're forcing the user to wire up two event handlers instead of only needing one. Ultimately the "previous/next" data point can be inferred from the delta with the previous value, so is easily obtainable if downstream apps needed it.

CraigCav:
We can probably leverage the `measure` callback inside `useCurrentPage`, since that is already wired up to only fire as a result of user interaction.

That makes this a pretty small  (`+6-5`) change to the codebase overall:

image

CraigCav (approved):

Issue #671, opened 2021-07-12
Enhancement: `EzField` for types text and textarea could include an `aria-invalid` and `aria-describedby` when error state exists


ez-alexfrazer:
**Is your feature request related to a problem? Please describe.**

Frequently, I will have to use this assertion:

https://github.com/testing-library/jest-dom#tobeinvalid

This is checking the `aria-invalid` property, so I would expect to see it within the input when its invalid

**Describe the solution you'd like**

I would expect the input to have both an `aria-invalid` and an `aria-describedby` pointing to the error when EzField is in an error state.

This would help with a lot of testing related problems. 

**Additional context**

Currently, it renders a tree like so:
![image](https://user-images.githubusercontent.com/77698413/125314802-84a96380-e304-11eb-868d-bb244a047ed3.png)

```html
<div class="sxs2lsm sxs2lsm4g5bi--labelSize-normal">
  <div>
    <div style="position: relative">
      <label
        id="control__2"
        for="control__1"
        class="
          sxhwbfs
          sxhwbfs09dbi--size-normal
          sxhwbfs03kze--error-true
          sxhwbfs2y18d--error-true
        "
      >
        Value
      </label>
    </div>
  </div>
  <div>
    <div style="position: relative">
      <input
        type="text"
        placeholder="Enter value"
        label="Value"
        id="control__1"
        name="control__1"
        aria-labelledby="control__2"
        class="
          sx4wg0o
          sx4wg0o03kze--error-true sx4wg0o03kze--touched-true
          sx4wg0onsprs--showInlineError-true
          sx4wg0o6i5lx--c2
        "
        value="asdf"
      />
      <div class="sxchvxj sxrqz6a sxiv9lz sxq6u3l">
        <svg viewBox="0 0 14 14" height="1rem" width="1rem" class="sxcothp">
          <path
[...]

PR #691, opened 2021-08-06
[DF-67] Add aria-invalid and aria-describedby to EzField

danidewitt:
## What did we change?
- Added aria-invalid and aria-describedBy to EzField

## Why are we doing this?
- Improve accessibility
- Improve testing
- Closes https://github.com/ezcater/recipe/issues/671

## Screenshot(s) / Gif(s):

Screen Shot 2021-08-06 at 11 20 50 AM

Screen Shot 2021-08-06 at 11 21 37 AM

## Checklist

### DIFF
--- a/src/components/EzField/EzField.tsx
+++ b/src/components/EzField/EzField.tsx
@@ -145,9 +145,14 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
   const active = focused || hovered;
   const showError = Boolean(touched && error);
   const labelType = isChoiceElement ? 'div' : 'label';
+  const errorMessageId = `errorMessage-${id}`;
+  const helperTextId = `helperText-${id}`;
   const errorMessage = (
     <div className={errorContainer()}>
-      <span className={clsx(errorCallout({active, inline: showInlineError}), arrow())}>
+      <span
+        id={errorMessageId}
+        className={clsx(errorCallout({active, inline: showInlineError}), arrow())}
+      >
         {error}
       </span>
     </div>
@@ -182,7 +187,11 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
             {!showInlineError && showError && errorMessage}
           </div>
         )}
-        {helperText && <div className={helper()}>{helperText}</div>}
+        {helperText && (
+          <div id={helperTextId} className={helper()}>
+            {helperText}
+          </div>
+        )}
         <div>
           <div style={relative}>
             <Input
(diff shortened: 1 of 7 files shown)


src/components/EzField/EzField.tsx
```diff
@@ -192,6 +201,8 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
                 id,
                 name: props.name || id,
                 'aria-labelledby': labelId,
+                'aria-invalid': showError,
+                'aria-describedby': `${showError ? errorMessageId : helperTextId || labelId}`,
```
ez-alexfrazer, line 205: what happens if you have both helper text _and_ an error message? Does only one show?

danidewitt: it will show the error message, you can see that in this [test](https://github.com/ezcater/recipe/pull/691/files#diff-0cb238eedf19cbb6b09e35857c0cc169cb4855104c48bf0d803296fbb80e4ffeR149), following this [spec](https://www.w3.org/TR/WCAG20-TECHS/ARIA21.html).

CraigCav: Should we consider changing this to support multiple, as `aria-describedby` supports multiple ids?

This group of accessibility consultants seems to think it's a good idea: https://www.tpgi.com/describing-aria-describedby/#:~:text=error_msg%20desc_group.

danidewitt: yes!

src/components/EzField/EzField.tsx
```diff
@@ -192,6 +201,8 @@ const EzField = forwardRef<HTMLElement, Props>((props, ref) => {
                 id,
                 name: props.name || id,
                 'aria-labelledby': labelId,
+                'aria-invalid': showError,
+                'aria-describedby': clsx(helperTextId || labelId, showError && errorMessageId),
```
CraigCav, line 205: This doesn't look quite right, but perhaps I'm missing something.

From what I can tell, `helperTextId` always has a value (`helperText-${id}`), so `labelId` will never be applied. I *think* that's probably OK, since `labelId` probably _shouldn't_ be applied to `aria-describedby`, since it is already used for `aria-labelledby`. That being said, code paths that aren't used (and we don't want to be used) should probably be removed.

It might also make sense to lead with the error message, since it provides the most relevant contextual information first.

Perhaps we want something like:
```suggestion
                'aria-describedby': clsx(showError && errorMessageId, helperText && helperTextId),
```

danidewitt: updated

CraigCav (approved):

Issue #695, opened 2021-08-31
Bug: Recipe 12 IE11 broken with Next.js 11.1


pgrippi:
## Describe the bug

First discovered in store-next

Something about next.js 11.1.X breaks recipe 12 in IE11. I tracked it down to the compiled `dist/esm/components/theme.config.js` - specifically the return statement in `mergeCss`:

```js
  return _objectSpread(_objectSpread({}, stitches), {}, {
    css: css
  });
```

When this is executed IE11 blows up, saying that it cannot iterate over a non-iterable object.

```
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
  at m (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:777893)
  at K (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:792108)
  at Q (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:790074)
  at Anonymous function (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:789277)
  at css (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:805038)
  at Anonymous function (https://static.cdn-ezcater.com/store-next/_next/static/chunks/pages/_app-1642d05ad448af6c0d53.js:1:807091)
  at n (https://static.cdn-ezcater.com/store-next/_next/static/chunks/webpack-d9fb22640d4ee5be6114.js:1:144)
```

This error happens in the `_app` component, which prevents next from booting and ends up showing a blank white page to IE11 users. This is blocking us from upgrading next to address a CVE.

## Steps To Reproduce

1. Add recipe to a next.js 11.1.X app and load the page in IE11

pgrippi:
Closed as recipe no longer supports IE11 and neither does the marketplace, so there is no need to fix this!

PR #699, opened 2021-09-08
Use NextJS to power the doc-site

CraigCav:
## What did we change?

- Created a new `/docs` directory, built with NextJS
- Added pages for
  - components list (i.e. `/components` route)
  - component details (i.e. `/components/{some-component-name}` route)
  - releases page
  - cookbook list + details
  - guides (contributing, getting started etc)
  - support (migration guides)
  - team
- Built new `CodeDemo` component leveraging build-time MDX, rather than needing a run-time babel/buble dependency
- Added google tag manager via `react-gtm-module` (to replace our use of `gatsby-plugin-google-tagmanager`)
- Reworked documented examples leveraging `react-router` (to avoid calling `require` within code examples)

## Why are we doing this?

A small number of our downstream applications use NextJS today, so using NextJS to power the doc-site provides a more representative example usage, than gatsby.

Aside from this, our gatsby site is rather slow to start up, which has a pretty significant impact on our cycle time. Rebuilding the docs gives us an opportunity to re-work some of the underpinnings to significantly reduce the start up time.

## Checklist (Still to do)

- [x] Add theme switcher to NextJS site
- [ ] Add search to NextJS site
- [ ] Re-work wrapper logic in `lib/mdx` into remark plugin (i.e. rewrites MDX to use wrap code snippets with demo/preview)
- [ ] Use local Recipe install for doc-site (rather than fixed Recipe version from npm)
- [x] Update build/release tooling to build NextJS site
- [x] Delete old gatsby code
- [ ] Wire up analytics to the new docs


docs/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "dependencies": {
+    "@ezcater/recipe": "0.0.0-canary-2021716142852",
+    "@recipe-ui/theme-marketplace": "^2.0.0",
+    "@types/react-gtm-module": "^2.0.1",
+    "@types/react-router-dom": "^5.1.8",
+    "glob": "^7.1.7",
+    "gray-matter": "^4.0.3",
+    "mdast-util-toc": "^5.1.0",
+    "mdx-bundler": "^5.2.1",
+    "next": "11.0.1",
```
pgrippi, line 22: did you get a chance to try next.js 1.1.2?

CraigCav: I didn't... but I will give it a spin now - I suspect  I'll [bump into the same issue as you](https://github.com/ezcater/recipe/issues/695), and if so, it makes sense to tackle that separately (this PR is already chunky).

Chunky - moto moto from Madagascar

CraigCav: FWIW, this PR as-is currently doesn't work for IE11 (despite still being on 11.0.1), without also needing additional [next-transpile-modules](https://www.npmjs.com/package/next-transpile-modules) configuration pointed at Recipe, since Recipe imports from `@react-stately` which isn't published to npm with ES5 compatible syntax. We have a ticket in Jira - `DF-77` - for tackling this.

Aside from Recipe, the doc-site also imports `playroom/utils`, which also isn't ES5 compatible. I would be happy enough to run this through `next-transpile-modules` for the short term (since this isn't a dependency of Recipe, and is only a dependency of the doc-site).

CraigCav:
Following up on [this discussion](https://github.com/ezcater/recipe/pull/699#discussion_r713110118), I wanted to add some more notes on IE11 support, since this is non-trivial due to some doc-site changes made in this PR.

Prior to this PR, the doc-site had a [run-time build step](https://github.com/ezcater/recipe/blob/c45c3ce4c838cf6e932c68fc3a3968ad7a2536eb/doc-site/src/components/Playground/Preview.js#L2) for rendering live previews of code examples. This build step was particularly heavy-weight, as it compiled _**static code examples**_ at runtime (using `buble` - similar to babel) in order to make the examples executable in the page. 

Since these code examples are static (i.e. we don't need live-editor support, as we use playroom for that), I decided to move this process to build time using [mdx-bundler](https://github.com/kentcdodds/mdx-bundler), allowing us to speed up the runtime and ship less code to the browser. The drawback of this approach (that I hadn't considered ahead of time 😳 ) was that mdx-bundler uses esbuild which has a minimum target language of ES6 - meaning that our code examples are transformed into modern JS that is incompatible with IE11.

This ultimately means we can't support IE11 without switching from away from using `mdx-bundler`, perhaps using something like [next-mdx-remote](https://github.com/hashicorp/next-mdx-remote) instead.

noranda:
Closing this PR. Recipe will be moving towards Storybook Docs in place of Next.JS for our next iteration of documentation. Thanks for your contribution!

PR #715, opened 2021-11-03
Next.js debugging

pgrippi:
## What did we change?

* Upgrade @ezcater/snitches to fix an issue with SVG style hydration
* Remove the custom Symbol polyfill in favor of relying on the consuming application to supply it

## Why are we doing this?

Hoping it will fix the issues we're seeing in IE11 when upgrading to next.js 11.1 & 12.0, as well as the issues ezmanage is seeing with their recipe 12 upgrade.

## Checklist


pgrippi:
Did not fix the issue - next.js 12.0 still dies with a white page and a SyntaxError with no stacktrace

PR #719, opened 2021-11-30
Update EzTimeInput for RHP related usage

onocy:
## What did we change?
- Added two optional props to `EzField`: 
	  - `displayAsNoon: <boolean>`: Replaces the "12:00 PM" label with "Noon"
	  - `focusLabel: <string>`: Determines which option is guaranteed to be within view when the dropdown opens

## Why are we doing this?
- We would like this behavior for the time dropdown on RHP to swap 12:00 PM with noon, and to make it so that noon is one of the first options seen when a user opens the dropdown. 

## Screenshot(s) / Gif(s):

Using the following code in the playroom produces the behavior below in which the 3:30 PM option is within the initial view when opening the dropdown. A developer can customize this prop to ensure that the field functions as intended. 

```js
{(() => {
  const [time, setTime] = React.useState('');
  return (
    <EzFormLayout>
      <EzField
        type="time"
        value={time}
        start="9:00 am"
        end="5:00 pm"
        step={30}
        label="Select delivery time"
        placeholder="Choose..."
        helperText="This is the time your food will be delivered."
        onChange={e => setTime(e.target.value)}
        focusLabel={'3:30 PM'}
        displayAsNoon
      />
    </EzFormLayout>
  );
})()}
```

https://user-images.githubusercontent.com/11084873/144135123-668c186b-e190-4784-b4e2-4355d0e97406.mp4

## Checklist

- [X] Provide test coverage for the changes made
- [X] Create a changeset (`npm run changeset`) with a summary of the changes


onocy:
I am not sure how to update the snapshot tests (I believe they run against the `.md` files?), but there are two currently that are failing that I'm not sure how to fix at this moment.

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +34,35 @@ const layout = theme.css({
...
   const date = dayjs().format(t('DATE_FORMAT'));
   const valueTime = dayjs(`${date} ${value}`);
   const valueTimeString = valueTime.format(t('TIME_FORMAT'));
+
   const {error, touched} = rest;
 
-  const options = useTimeRangeOptions({start, end, step});
+  const options = useTimeRangeOptions({start, end, step, displayAsNoon});
+
+  const selectOptions = useMemo(
```
pgrippi, line 56: I don't think this `useMemo` is functioning like you want, `options` is created as a fresh array every time this component renders, so this memo won't actually cache anything

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +34,35 @@ const layout = theme.css({
...
+    () =>
+      options.map(option =>
+        option === 'Noon' && displayAsNoon
+          ? {label: option, value: '12:00 PM'}
+          : {label: option, value: option}
```
pgrippi, line 61: You have this logic twice, once here and once in `useTimeRangeOptions`. I think you could remove your changes to `useTimeRangeOptions` and just have this:

```js
options.map(option => ({
  label: option === '12:00 PM' && displayAsNoon ? 'Noon' : option,
  value: option
}))
```

src/components/EzField/EzListBox.tsx
```diff
@@ -100,13 +100,14 @@ const EzListBox = (props, ref) => {
     /* Note: lint doesn't detect the keyboard handler is on the input element, not the list */
     const current = selectionManager.selectedKey === option.key;
     const selected = selectionManager.focusedKey === option.key;
+    const shouldReceiveFocus = props.focusLabel === option.label;
```
pgrippi, line 103: Does this focus ever need to be cleared? Your check below is `selected || shouldReceiveFocus`, which makes me think that if you pass `focusLabel="3:00 PM"` and you select 9 AM, it will still keep the focus on 3:00 PM as it comes after and they will both pass the same ref (9 AM is `selected` and 3 PM is `focusLabel`)

src/components/EzField/EzTimeInput.tsx
onocy: sounds good, thank you

src/components/EzField/EzListBox.tsx
onocy: Yes you're right. After a value is entered, the focusLabel should not affect the focus.

src/components/EzField/EzTimeInput.tsx
onocy: Shoot, you're right. I'll change.

src/components/EzField/EzTimeInput.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useMemo} from 'react';
```
pgrippi, line 1: can ✂️ this now

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +29,32 @@ const layout = theme.css({
...
 
   const date = dayjs().format(t('DATE_FORMAT'));
   const valueTime = dayjs(`${date} ${value}`);
   const valueTimeString = valueTime.format(t('TIME_FORMAT'));
+
   const {error, touched} = rest;
 
   const options = useTimeRangeOptions({start, end, step});
+
+  const selectOptions = () => {
```
pgrippi, line 51: This is more of a styling thing, but I don't think this needs to be a function. I like pulling it out of the JSX below, but I think you can just do `const selectOptions = options.map(() => ...)`

pgrippi (approved):

.changeset/big-yaks-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': minor
+---
+
+Added optional props to EzField for time inputs
```
morgannegagne, line 5: 🎱  might be nice to include what the optional props are

src/components/EzField/EzField.types.ts
```diff
@@ -53,6 +53,7 @@ type TimeInputProps = {
   start: string;
   end: string;
   step?: number;
+  displayAsNoon?: boolean;
```
morgannegagne, line 56: do we need to add the `focusLabel` here too?

.changeset/big-yaks-vanish.md
onocy: Sure, I wasn't sure how much detail to add in this but I can do this np

src/components/EzField/EzField.types.ts
onocy: Hmm, I had thought I did but I can add.

src/components/EzField/EzField.md
```diff
@@ -551,6 +551,11 @@ Allows the user to pick a time from a select dropdown. Use `type="time"` to enab
 
 The steps that are supported are: `5`, `10`, `15`, `20`, `30`, `60`
 
+Optional Props:
+
+- `displayAsNoon: <boolean>`: Replaces the "12:00 PM" label with "Noon"
```
evan-007, line 556: ❓ is this file used in the doc site?

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +29,30 @@ const layout = theme.css({
...
 
   const date = dayjs().format(t('DATE_FORMAT'));
   const valueTime = dayjs(`${date} ${value}`);
   const valueTimeString = valueTime.format(t('TIME_FORMAT'));
+
   const {error, touched} = rest;
 
   const options = useTimeRangeOptions({start, end, step});
+
+  const selectOptions = options.map(option => ({
```
evan-007, line 51: should we `useMemo` on `selectOptions`? it wasn't before, but maybe it's worth doing

src/components/EzField/EzListBox.tsx
```diff
@@ -100,13 +100,15 @@ const EzListBox = (props, ref) => {
     /* Note: lint doesn't detect the keyboard handler is on the input element, not the list */
     const current = selectionManager.selectedKey === option.key;
     const selected = selectionManager.focusedKey === option.key;
+    const shouldReceiveFocus =
+      props.focusLabel === option.label && selectionManager.selectedKey === null;
```
evan-007, line 104: ❓ what does this mean? `selectionManager.selectedKey === null;` the user has not selected an option? so when `props.focusLabel` is present, the app that renders this component doesn't need to manage that prop as the user selects something?

so we don't need to do
```
() => {
  const [choice, setChoice] = React.useState(null);
  const focusLabel = choice === null ? 'All Time' : undefined;  // is this line needed?
  return (
    <EzFormLayout>
      <EzField
        type="select"
        label="Select dropdown"
        placeholder="Choose..."
        options={[
          {label: 'All Upcoming', value: 'upcoming'},
          {label: 'Today', value: 'today'},
          {label: 'Tomorrow', value: 'tomorrow'},
          {label: 'All Time', value: 'all'},
          {label: 'Yesterday', value: 'yesterday'},
          {label: 'Last 7 Days', value: 'week'},
          {label: 'This Month', value: 'month'},
        ]}
        value={choice}
        onSelectionChange={setChoice}
      />
    </EzFormLayout>
  );
};
```

evan-007 (commented):
LGTM

is there a way to test `focusLabel` or is the snapshot `__tests__/__image_snapshots__/chrome-desktop-ezfield-time-focus-label-snap.png` doing that?

src/components/EzField/EzField.md
onocy: Yes!
Screen Shot 2021-12-01 at 3 33 59 PM

src/components/EzField/EzTimeInput.tsx
onocy: I had this before, but as Peter mentioned [here](https://github.com/ezcater/recipe/pull/719#discussion_r760183112) `options` is created as a fresh array every time this component renders, so the memo won't actually cache anything.

src/components/EzField/EzListBox.tsx
onocy: Essentially `selectionManager.selectedKey === null` here is just to prevent the focusLabel from _always_ receiving focus. We want it to receive focus when the dropdown is open, but after an option is selected (and the dropdown is reopened) the selected option should receive focus priority (the `focusLabel` shouldn't matter anymore, it's just relevant before you select something).

onocy:
@evan-007 TBH the snapshot there isn't a very useful test. I couldn't find any tests for the existing focus behavior, so I only proceeded from a manual test (I also wasn't sure how much value it would add). 

I suppose I could have a test that opens the dropdown initially with a `focusLabel` on one of the last options and see if it's visible? 

If you have thoughts on this I'm all👂 s.

src/components/EzField/EzListBox.tsx
evan-007: 👍 is it worth adding that as a comment since we probably can't unit test that very easily?

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +29,30 @@ const layout = theme.css({
...
 });
 
-const EzTimeInput = ({start, end, step = 60, value, ...rest}) => {
+const EzTimeInput = ({
+  start,
+  end,
+  step = 60,
+  value,
+  displayAsNoon = false,
+  focusLabel = '',
```
danidewitt, line 38: was `''` the previous default?

src/components/EzField/EzTimeInput.tsx
```diff
@@ -29,15 +29,30 @@ const layout = theme.css({
...
   const date = dayjs().format(t('DATE_FORMAT'));
   const valueTime = dayjs(`${date} ${value}`);
   const valueTimeString = valueTime.format(t('TIME_FORMAT'));
+
   const {error, touched} = rest;
 
   const options = useTimeRangeOptions({start, end, step});
+
+  const selectOptions = options.map(option => ({
+    label: option === '12:00 PM' && displayAsNoon ? 'Noon' : option,
```
danidewitt, line 52: have we thought about if this breaks anything related to internationalization?

src/components/EzField/__tests__/EzField.time.test.tsx
```diff
@@ -118,4 +130,16 @@ describe('EzField', () => {
     expect(options[7]).toHaveTextContent('4:00 PM');
     expect(options[8]).toHaveTextContent('5:00 PM');
   };
+
+  const expectHourlyOptionsFrom9to5ExceptNoon = options => {
```
danidewitt, line 134: thank you for adding tests 🎉

src/components/EzField/EzTimeInput.tsx
danidewitt: oh this is the completely new prop, disregard

.changeset/big-yaks-vanish.md
```diff
@@ -0,0 +1,8 @@
...
+
```
danidewitt, line 6: can you add the type of change here please? this probably wasn't super obvious but we are following [this commit message convention](https://github.com/ezcater/recipe/blob/main/doc-site/src/pages/guides/versioning-and-changelog.md)

danidewitt:
@onocy left a couple comments. the only other thing aside from those comments would rebasing your commits down with the respective changeset commits (like [this](https://github.com/ezcater/recipe/pull/690/commits))

.changeset/big-yaks-vanish.md
danidewitt: this also needs to be better documented, we're trying to rebase changes with respective changeset commits like [this](https://github.com/ezcater/recipe/pull/690/commits)

src/components/EzField/EzTimeInput.tsx
onocy: I added a translation for this in the most recent commit

danidewitt:
@onocy did the commits not get rebased?

PR #728, opened 2021-12-15
[DF-288] Autohide EzFlashMessage

danidewitt:
## What did we change?
- Added `autohide`, `autohideDuration`, and `onAutohide` to `EzFlashMessage`

## Why are we doing this?
- To allow flash messages to disappear automatically after a certain amount of time, as requested by designers

## Screenshot(s) / Gif(s):

https://user-images.githubusercontent.com/3790037/146276490-c0b43a1b-5d4f-4652-9154-43e764dbd361.mov

## Checklist


src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -85,6 +88,23 @@ const icons = {
...
+  useEffect(() => {
+    if (!props.autohide) return;
+
+    const autohideTimer = setTimeout(() => {
+      if (props.onAutohide) props.onAutohide();
+      setShow(false);
+    }, props.autohideDuration || 5000);
+
+    // eslint-disable-next-line consistent-return
+    return () => clearTimeout(autohideTimer);
```
danidewitt, line 103: i need to test a few extra times: 
1.  what it looks like to leave and come back, does the timeout blow up?
2. what does it look like with 100 autohide messages at once?

there's a very small chance we should also just build a snackbar component for this.

danidewitt: we're not going to build a snackbar now due to time constraints

jbelang:
Also going to CC @spdonegan1 and @kevinmook because ezManage will be using this in conjunction with its flash message library and I think these two have the most context on how this will be compatible with that.

src/components/EzFlashMessage/EzFlashMessage.tsx
jbelang: The two things you called out for extra testing are good ideas. When I briefly attempted to roll my own version of this, those were two areas where I ran into problems. A third was that, even when the message was hidden, components around it were still slightly shifted from where they otherwise would have been if the message was never there.

jbelang (approved):
This looks good to me, pending the investigations you called out about edge cases.

kevinmook (approved):
LGTM!

WRT the ezmanage flash message library, I think this should work well with that. It's been a while since I thought about that flash message library but iirc it's also not too complex. Let me know if I can help with anything!

src/components/EzFlashMessage/EzFlashMessage.tsx
danidewitt: this is one way to test the components shifting around it and having a bunch on the screen (i can't share the actual playroom because local, but if you pull it down you can paste this in to a local playroom):

```
{(() => {
  return (
    <EzPage>
      <EzFlashMessage use="success">
        <p>1</p>
      </EzFlashMessage>
      <EzFlashMessage use="success">
        <p>2</p>
      </EzFlashMessage>
      <EzFlashMessage use="success">
        <p>3</p>
      </EzFlashMessage>
      <EzFlashMessage
        use="success"
        headline="We saved that for you."
        autohide
        autohideDuration={2000}
        onAutohide={() => console.log('Flash message hidden.')}
      >
        <p>
          4
        </p>
      </EzFlashMessage>
      {[...Array(100).keys()].map(index => {
        return (
          <EzFlashMessage use="success">
            <p>{index+4+1}</p>
          </EzFlashMessage>
        );
      })}
    </EzPage>
  );
})()}
```

danidewitt: @jbelang how many are on a screen at once, at the top end of the range, typically?

danidewitt: @jbelang and did they shift horizontally or vertically, or both?

danidewitt: i don't see the timeouts blowing up when i leave the page - i think it's cleaning them up properly

jbelang: I had been playing with a component outside of Recipe, so I don't know that any of it was too relevant. 

For our use case, the absolute most that could be on screen at once would be around 50, since that's the most orders we display that they could click through. More reasonably, my gut says it'd be 5 - 10.

When I had even one show up and then hide itself, the message properly vertically aligned relative to everything else on the page, but then, when it hid, the rest of the page was still a bit vertically shifted down. But I think that was more an issue with what I was trying to do than anything I'm concerned about with Recipe.

jbelang:
Do we have a rough idea on when this will land in a Recipe release?

src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -85,6 +88,23 @@ const icons = {
...
 const EzFlashMessage: React.FC<FlashMessageProps> = initProps => {
   const {props} = flashMessage(initProps);
+
+  const [show, setShow] = React.useState(true);
+
+  useEffect(() => {
+    if (!props.autohide) return;
+
+    const autohideTimer = setTimeout(() => {
+      if (props.onAutohide) props.onAutohide();
```
ez-alexfrazer, line 98: I think you can do `props.onAutohide?.()` as shorthand?

danidewitt:
tests passing

Screen Shot 2022-02-25 at 7 00 59 PM

PR #730, opened 2021-12-18
Fix AutoSuggest/ComboBox onBlur bug

### LINKED ISSUE #713: Bug: EzField `autosuggest` doesn't work when selecting an item after filtering
## Describe the bug

When you type a value in an `EzField` with type `autosuggest` and then click one of the selected options, it does not get selected and the component resets the filter and shows all options again.

## Steps To Reproduce

1. Go to https://ezcater.github.io/recipe/components/ez-field#autosuggest-list
2. In the component type `al` and then try to click the `Tallahassee, Florida` option
3. Notice that the field gets cleared and the selection is not made

## Expected behavior

Selecting a value after filtering should work.

## Additional context

I've noticed a couple more things:

* The field probably should have auto-suggestions turned off, as the browser auto suggestion interferes with it (this is on Chrome latest version):
image

* If you type something in the component, switch to a different tab on the browser and come back, whatever you typed disappears

setphen:
When a `mouseDown` occurs within a filtered results list, the ComboBox `onBlur` event is fired, triggering a render, before a `onClick` can be registered. This change checks if the popover is still present to determine if the `onBlur` happens inside or outside of the list box.

Fixes #713 

## Screenshot(s) / Gif(s):

https://user-images.githubusercontent.com/3891632/146647545-e2a6eaf8-519a-4b19-9045-994c6cf50528.mov

### DIFF
--- a/.changeset/neat-dodos-applaud.md
+++ b/.changeset/neat-dodos-applaud.md
@@ -1,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+Fix: AutoSuggest onBlur bug when clicking item within filtered collection
--- a/src/components/EzField/useComboBox.ts
+++ b/src/components/EzField/useComboBox.ts
@@ -89,8 +89,8 @@ export function useComboBox(props, state): ComboBoxAria {
     // call the user defined onBlur (if provided)
     props.onBlur?.(e);
 
-    // clear the input if no value was selected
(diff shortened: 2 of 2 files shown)


noranda (approved):
All tests pass and bug is fixed! Thanks!

pgrippi (approved):

PR #731, opened 2022-01-04
Export constants used in theme.config.js [PAC- 1492]

onocy:
## What did we change?
- Added a base theme that contains constants used when the recipe theme is defined initially.
- Added a marketplace that imports the base theme and contains potential marketplace specific overrides.
- By adding these themes, this PR pulls constants out of the `theme.config.js` in a way that we can consume those constants in downstream services. 
    - Without adding stitches to our other apps, the easiest way to pull in the values is to export the constants as objects and then import these objects within our services.
    - Once imported into our services, we can define an [emotion theme](https://emotion.sh/docs/theming) with the imported objects or have inline styles referencing the imported values as follows: 
    ```js
    import {themes} from '@ezcater/recipe';
    
    ...
    
    //  emotion theme 
    
    const SomeText = styled.div`
	    color: ${props => props.theme.colors.blue600};
	`;
	
	render(
	    <ThemeProvider theme={themes.marketplace}>
	      <SomeText>Text</SomeText>
	    </ThemeProvider>
  	)

	...
  	
    // vs. inline styles
   	 const SomeText = styled.div`
	    color: ${themes.marketplace.colors.blue600};
	`;

	render(<SomeText>Text</SomeText>)
    ```
   
## Why are we doing this?
- We have global styles files defined in our apps, but would like to work towards having recipe become the source of truth for constants (relating to color, font-size, margin, padding, etc.)
- We used to export a large amount of css in the [legacy-theme](https://github.com/ezcater/recipe/blob/main/packages/%40recipe-ui/legacy-theme/src/index.ts) which used a similar approach of exporting objects. I'm not sure what the push away from that was determined by. 

## Screenshot(s) / Gif(s):

Quick test in store-next after linking with recipe and overriding the button styles with the marketplace theme object (button becomes purple since the marketplace theme override color for blue600 is purple): 
Screen Shot 2022-01-04 at 5 02 40 PM

## Checklist

- [X] Create a changeset (`npm run changeset`) with a summary of the changes


src/themes/base.ts
```diff
@@ -0,0 +1,126 @@
...
+  },
+  fontSizes: {
+    75: '12px',
+    100: '14px',
+    200: '16px',
+    300: '18px',
+    450: '21px',
+    800: '32px',
+    // aliases
+    text: 'var(--recipe-base-font-size, $200)',
```
onocy, line 78: There are variables such as the one above that are used in association with stiches, so I believe it would break in the marketplace if we tried using this particular style- so we'd have to find a workaround for these cases (maybe simplest to have overrides in `theme.config.js` that add these?)

onocy:
❔ A couple of open questions here: 

1. If we are going through with this approach- Is this the proper location for the theme? (in `src/themes`)
2. Does pulling the `base` theme object into `theme.config.js` affect recipe in any way or any plans for the future?

src/themes/base.ts
evan-007: some of these like https://github.com/ezcater/recipe/pull/731/files#diff-2c07671211f42131b11ec88a7a4dce1d5a969980996c0dc3ed65cc8bbe09bf7aR49-R57 we could probably just pull into js variables and interpolate them here, but i'm not sure how that will play with stitches

evan-007:
this came out pretty well! an approach like this makes sense to me from an app developer point of view, i can't comment on if it's the right thing for recipe tho.

i do like the theming approach w/emotion more than the legacy theme's approach of exporting a bunch of different objects like how store-react does styles.

morgannegagne (approved):
LGTM!

src/themes/base.ts
pgrippi: We could split apart the base theme from the semantic theme - so `green200` would live in this file but `positiveBg` wouldn't, for example:

```js
const stitches = createCss({
  ...root,
  ...base,
  colors: {
    ...base.colors,
    positiveBg: '$green200',
  }
});
```

That would get around us needing to support the `var(--)` and `$colorName` interpolations, I would think that those semantic styles are recipe-specific and probably not something a consuming application would want to know

pgrippi (dismissed):
I think this is a great starting point, and exposes a path forward for the store projects to consume these variables without specific code for stitches or emotion

I wonder if there is some way we could also expose the `media` keys in the theme, I think that is the only thing missing from this first pass!

onocy:
I resent out for review despite getting some approvals already just to be sure that folks are OK with the recent updates (pulling the recipe/stiches vars out of the base theme and adding the media queries to the base theme). 

If we are OK with this PR i'm going to add a recipe feature request for Evan's idea related to having a theme page that maybe we can take a look at in the future.

src/components/theme.config.ts
```diff
@@ -21,55 +22,11 @@ const root = {root: createRoot()};
...
-      yellow200: '#fcf6e5',
-      yellow500: '#f4d689',
-      yellow600: '#e9a801',
-      yellow700: '#926a00',
-
-      transparent: 'transparent',
-
+      ...base.theme.colors,
       // semantic
       positiveBg: '$green200',
```
morgannegagne, line 31: 🎓 ❓  Why are these values not pulled into the base theme?

morgannegagne (approved):

src/components/theme.config.ts
onocy: All of these values use a concept from stiches called [tokens ](https://stitches.dev/docs/tokens) which are basically variables. 

If we used these constants out of a stiches context the mapping wouldn't translate correctly, causing some issues.
 
Ultimately these are essentially aliases for existing values, and while they are used in recipe, we figured it was fine to just take it out of the base theme which will be used elsewhere (like marketplace)

morgannegagne: cool that makes sense! thanks for the explanation

src/themes/base.ts
```diff
@@ -0,0 +1,114 @@
...
+      red950: '#972f2f',
+      teal200: '#effaf8',
+      teal600: '#1bbc9b',
+      teal700: '#008066',
+      white: '#fff',
+      yellow200: '#fcf6e5',
+      yellow500: '#f4d689',
+      yellow600: '#e9a801',
+      yellow700: '#926a00',
+
```
evan-007, line 46: ❓ why the break here?

src/themes/base.ts
```diff
@@ -0,0 +1,114 @@
...
+    lineHeights: {
+      1: 1,
+      // aliases
+      snug: 1.35,
+      tight: 1.25,
+    },
+    radii: {
+      // sizes
+      regular: '4px',
+
```
evan-007, line 73: and here?

evan-007 (approved):

danidewitt:
@onocy thank you for this PR and apologies for the delay! this is an awesome submission and informed our decisions about mui theming! we're going to close this and make these tokens available to you in an emotion theme (which should be easier to merge with theming in your application)

PR #737, opened 2022-01-26
Add .tool-versions with working node version

evan-007:
## What did we change?

Add .tool-versions with working node version

## Why are we doing this?

I had trouble getting `npm run start` to run locally, because my system node version was 16.0.0. I also tried 17.x and couldn't get it to work.

14.x is what the [github actions use](https://github.com/ezcater/recipe/blob/3ce549de648ca341c380cb525826fdc72e8bf9ca/.github/workflows/release.yml#L19-L22) so maybe we should specify the version in `.tool-versions` since ezcater is standardizing around asdf.

## Checklist


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 14.18.3
```
evan-007, line 1: this is the latest version, it would be nice to do `14.x` like gh actions uses but i'm not sure if .tool-versions supports that.

morgannegagne (approved):

pgrippi (approved):

pgrippi (approved):

onocy (approved):

noranda (approved):
Looks good! All tests pass.

Screen Shot 2022-02-24 at 9 57 46 AM

PR #742, opened 2022-01-27
feat: all onFocus callback events to  EzField type="date"

evan-007:
## What did we change?

allow EzField type="date" to take onFocus + onFocusCapture props

## Why are we doing this?

We want a way to hide the native keyboard on mobile devices when the datepicker is used, and i think the only way to do that is to blur the input as soon as it has focus. See https://github.com/ezcater/store-next/pull/1897 for example usage in one of our apps.

## Checklist


src/components/EzField/EzDateInput.tsx
```diff
@@ -99,6 +99,7 @@ const EzDateInput = ({
     placeholder,
     error: props.error,
     touched: props.touched,
+    onFocus: props.onFocus,
```
evan-007, line 102: it's a little strange to me that all EzFields don't automatically support events like `onFocus` `onTouchStart` `onTouchEnd` etc. or maybe `EzDateInput` is the only weird one because it passes only certain props to the `<input>`. I'm not sure if I should add the others here or if that's part of a larger discussion

morgannegagne (approved):
I like this solution! super flexible

src/components/EzField/__tests__/EzField.date.test.tsx
```diff
@@ -291,5 +291,23 @@ describe('EzField', () => {
 
       expect(screen.getByRole('dialog')).toContainElement(document.activeElement as HTMLElement);
     });
+    it('should call props onFocusCapture and onFocus', () => {
```
danidewitt, line 294: 👍

src/components/EzField/EzDateInput.tsx
danidewitt: 🤔 i'm not entirely sure either - maybe add what you need now and create a ticket to investigate the rest

danidewitt (approved):

onocy (approved):

morgannegagne:
All tests pass locally
Screen Shot 2022-02-08 at 10 11 50 AM

danidewitt:
LGTM!

PR #761, opened 2022-02-07
Allow option for white background on EzPage

maggiedbaker:
## What did we change?
- Add an optional boolean `backgroundColor` prop to `EzPage`, which accepts the values `'white'` and `'gray'`
- Add a `backgroundColor` variant that handles the background color styling depending on the value of the `backgroundColor` prop. The default `backgroundColor` variant is `'gray'`. 

## Why are we doing this?
- So that we have the option of white backgrounds on pages, specifically in ezManage, and can also expand the background colors we allow.

## Screenshot(s) / Gif(s):
Feb-17-2022 09-53-28

## Checklist


ez-alexfrazer:
I think the only problem with this is that it doesn't leave a lot of room for extension. If we had to allow more colors, it would be a major release. Maybe something like this?

```ts
interface Props {
  backgroundColor?: 'white';
}
```

if not provided it will be gray, the only "changeable" option would be `'white'`

```ts
<EzPage backgroundColor="white">
</EzPage>
```

maggiedbaker:
> I think the only problem with this is that it doesn't leave a lot of room for extension. If we had to allow more colors, it would be a major release. Maybe something like this?
> 
> ```ts
> interface Props {
>   backgroundColor?: 'white' | 'gray';
> }
> ```
> 
> if not provided it would just default to gray. Benefit is that if you decide to expand it you can
> 
> ```tsx
> interface Props {
>   backgroundColor?: 'white' | 'gray.100' | 'gray.200' | 'gray.300' /** ... */ 'gray.900';
> }
> ```

Ah ok - I like this idea! Thank you.

src/components/EzPage/EzPage.theme.config.ts
```diff
@@ -4,6 +4,7 @@ export default mergeCss({
   theme: {
     colors: {
       'page-bg': '$gray200',
+      'page-bg-white': 'white',
```
danidewitt, line 7: very small: '$white'

danidewitt:
@maggiedbaker this looks really good - _especially_ after you made it more extensible. i left a small comment about the variable, but otherwise, let's coordinate on merge timing! 🎉

src/components/EzPage/EzPage.theme.config.ts
maggiedbaker: good catch! Updated.

danidewitt (approved):
🎉

danidewitt:
@maggiedbaker i made a couple very tiny edits to the docs and removed the extra test snapshots

visual tests are passing too
Screen Shot 2022-02-25 at 5 54 38 PM

PR #732, opened 2022-01-19
Enable Tree-shaking in consuming applications

pgrippi:
## What did we change?

This adds `sideEffects` to the root package.json, informing webpack which files can be safely tree-shaken if left unused. It also updates the `@recipe-ui` packages whose sideEffects listed CSS files that no longer exist

## Why are we doing this?

Now that IE11 support has been dropped I am working on moderizing the store-next build a bit. We noticed awhile back that recipe was not correctly tree shaking unused components, but we chaulked it up to the transpiling of the recipe library that was confusing webpack.

Once I removed the transpiliation step I noticed that recipe was still including the full bundle. Then I realized that we didn't have `sideEffects` in the root package.json, which was causing webpack to include all of recipe regardless of the components being used.

Build size before:

```
Page                                          Size     First Load JS
┌ λ /                                         320 B           351 kB
├   /_app                                     0 B             351 kB
├ λ /404                                      1.92 kB         414 kB
├ λ /500                                      354 B           351 kB
├ λ /catering-menu/[catererUrlId]             655 B           489 kB
├ λ /catering-menu/[catererUrlId]/[date]      662 B           489 kB
├ λ /catering-menu/pvt/[catererUrlId]         659 B           489 kB
├ λ /catering-menu/pvt/[catererUrlId]/[date]  665 B           489 kB
├ λ /catering/[catererUrlId]/packages         1.05 kB         430 kB
├ λ /catering/search/[orderId]                31.9 kB         527 kB
├ λ /feature-preview/nextjs                   581 B           408 kB
├ λ /health_check                             348 B           351 kB
└ λ /start-new-order                          23.5 kB         478 kB
+ First Load JS shared by all                 351 kB
  ├ chunks/framework.9b0cef.js                42.6 kB
  ├ chunks/main.1be061.js                     25 kB
  ├ chunks/pages/_app.6cbabd.js               281 kB
  ├ chunks/webpack.44d2c4.js                  2 kB
  └ css/db888522e9ce88d0b6c3.css              4.65 kB
```

Build size after:

```
Page                                          Size     First Load JS
┌ λ /                                         323 B           335 kB
├   /_app                                     0 B             335 kB
├ λ /404                                      1.66 kB         404 kB
├ λ /500                                      362 B           335 kB
├ λ /catering-menu/[catererUrlId]             665 B           483 kB
├ λ /catering-menu/[catererUrlId]/[date]      669 B           483 kB
├ λ /catering-menu/pvt/[catererUrlId]         668 B           483 kB
[...]

PR #762, opened 2022-02-14
feat: EzIcon component

danidewitt:
## What did we change?
- Added an `EzIcon` component that can accept icons from: ezCater icons, Font Awesome (free and pro), and SVGs. The components use mui `SvgIcon` and react-fontaweome `FontAwesomeIcon`. I have separated the interface `EzIcon` from the implementations (`EzIconMui`, `EzIconEzCater`, and `EzIconFontAwesome`) so that we can easily add more types of icons over time, while continuing to use the same interface and props. The icon is accessible using the `title` prop on `EzIcon`. Only the doc-site imports icons for documentation examples only, Recipe itself does not import icon packages. This reasoning is two-fold: performance and licensing. We cannot distribute paid pro icons as an open source project. 
- This introduces the usage of `mui` and `fontawesome`, and reintroduces `emotion`, which we should evaluate in the downstream applications before shipping. The decision to introduce these was vetted by the Recipe Steering Committee.

## Why are we doing this?
- Icons are used everywhere, we need to be able to use them on their own and within other components

## Screenshot(s) / Gif(s):
Screen Shot 2022-02-14 at 3 16 08 PM
Screen Shot 2022-02-14 at 3 16 29 PM
Screen Shot 2022-02-14 at 3 16 38 PM
Screen Shot 2022-02-14 at 3 16 43 PM
Screen Shot 2022-02-14 at 3 16 47 PM
Screen Shot 2022-02-14 at 3 16 52 PM
Screen Shot 2022-02-14 at 3 16 55 PM
Screen Shot 2022-02-14 at 3 18 01 PM

## Checklist


danidewitt:
tests passing locally
Screen Shot 2022-02-14 at 5 05 01 PM

ez-alexfrazer:
Will this allow us to use pro icons from fontawesome? I have found cases where I needed them

src/components/EzIcon/EzIcon.md
```diff
@@ -0,0 +1,137 @@
...
+        <EzTextStyle>Order fresh coffee from ezCater</EzTextStyle>
+      </EzLayout>
+    </EzPage>
+  );
+};
+```
+
+### Semantic icons
+
+Semantic icons communicate meaning. Semantic icons can be used on their own without text, as abbreviations, or in the context of interactive controls like buttons. For semantic icons, the meaning needs to be read by assistive devices. If an icon is semantic, include the `title` prop with an meaningful value to `EzIcon`.
```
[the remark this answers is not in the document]
danidewitt: good catch! thanks

danidewitt:
local tests still passing

morgannegagne (approved):
LGTM!

danidewitt:
@maggiedbaker did you try the canary release to see if it causes any dependency or other issues in your app?

ez-alexfrazer (approved):

maggiedbaker:
> @maggiedbaker did you try the canary release to see if it causes any dependency or other issues in your app?

Yes, I tried it locally in ezManage and first ran into `Cannot find module 'react/jsx-runtime'` which I think is due to Emotion 11 using [JSX Transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), which our version of React - 16.13.1 - doesn't support.

However, looks like the next version of react, [React 16.14.0](https://github.com/facebook/react/blob/main/CHANGELOG.md#16140-october-14-2020), was released just for this issue -- the only update in the changelog for that upgrade is support for JSX Transform. Seems like a pretty safe upgrade, so I've tried that and it appears to be working! Pushing up to dev1 this morning for some further investigation.

Does that React upgrade / JSX transform issue set off any red flags for anyone?

.changeset/rare-glasses-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': minor
```
maggiedbaker, line 2: Will this need to be a major release since it will require a bump of React to at least `16.14.0` in many apps?

maggiedbaker (approved):
This is awesome! I believe that apps will need to have a React version >= `16.14.0` to use this, but it looks like that should be a pretty easy bump for our apps that need it. We'll want to consider that for our release version though -- this might need to be a major release.

spdonegan1:
One thing I noticed is a pretty large bundle size increase on the ezmanage side, looking at the build logs from Maggie's test branch the main vendor bundle went from ~425kb to ~465kb (gzip size) of which it is all coming from recipe, seems like the added mui / fortawesome / emotion packages. I don't consider this a blocker from ezmanage perspective (i'll be sad but I'll get over it), but I don't know if any of the store folks want to test it / chime in to make sure this doesn't end up being a blocker for them.

pgrippi:
> One thing I noticed is a pretty large bundle size increase on the ezmanage side, looking at the build logs from Maggie's test branch the main vendor bundle went from ~425kb to ~465kb (gzip size) of which it is all coming from recipe, seems like the added mui / fortawesome / emotion packages. I don't consider this a blocker from ezmanage perspective (i'll be sad but I'll get over it), but I don't know if any of the store folks want to test it / chime in to make sure this doesn't end up being a blocker for them.

I think the increase in size is fine, and to be expected as we transition to using MUI under the hood.

I do know that one reason for the increase is because tree shaking isn't currently working in webpack, I have a PR open that fixes that: https://github.com/ezcater/recipe/pull/732, I don't know a good way to test both the changes here and the changes in that PR, but it might be worth a try to combine the two and see if that affects the build size in a positive way

spdonegan1 (approved):
LGTM pending Maggie's comment around the react peer dep / how to version that change.

.changeset/rare-glasses-grow.md
danidewitt: That's an interesting distinction - that thread definitely makes me lean towards major now. I'll discuss again with devex, thanks for linking that.

danidewitt:
visual tests still passing
Screen Shot 2022-02-24 at 10 05 34 AM

danidewitt:
tests passing

Screen Shot 2022-02-25 at 6 06 24 PM

PR #763, opened 2022-02-15
feat: EzRating component

danidewitt:
## What did we change?
- Added an icon rating component `EzRating`

## Why are we doing this?
- To be able to create a star rating component (flexible enough to use with other icons, too)

## Screenshot(s) / Gif(s):
Screen Shot 2022-02-15 at 1 24 07 PM
Screen Shot 2022-02-15 at 1 24 24 PM
Screen Shot 2022-02-15 at 1 24 35 PM

## Checklist


src/components/EzRating/Implementations/EzRatingCustom/EzRatingCustom.tsx
```diff
@@ -0,0 +1,40 @@
...
+  size,
+  color,
+}) => {
+  const valueRoundedToClosestHalfInteger = Math.round(value * 2) / 2;
+  const fullIconCount = Math.floor(valueRoundedToClosestHalfInteger);
+  const halfIconCount = Math.round(valueRoundedToClosestHalfInteger - fullIconCount);
+  const emptyIconCount = max - Math.ceil(valueRoundedToClosestHalfInteger);
+
+  return (
+    <span role="img" aria-label={label}>
```
ez-alexfrazer, line 21: 🎱 would this be more accessible as a [meter](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter) element? It also allows for some more granular control of [styling](https://css-tricks.com/html5-meter-element/)

src/components/EzRating/EzRating.md
```diff
@@ -0,0 +1,101 @@
...
+category: Data
+path: '/components/ez-rating'
+tags: ['rating']
+---
+
+The rating component is used to indicate how something has been assessed based on quality, standards, or performance.
+
+---
+
+## Examples
```
morgannegagne, line 13: Is it worth having an example with a `label` prop?

morgannegagne (approved):

src/components/EzRating/Implementations/EzRatingCustom/EzRatingCustom.tsx
ez-alexfrazer, line 21: I think the accessibility on this might be unclear, would a [`meter`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter) component make sense in this case?

> The <meter> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents either a scalar value within a known range or a fractional value.

src/components/EzRating/EzRating.types.ts
```diff
@@ -0,0 +1,17 @@
...
+export type EzRatingColors = EzIconColors;
+export type EzRatingSizes = EzIconSizes;
+
+export interface EzRatingProps {
+  max: number;
+  value: number;
+  label: string;
+  emptyIcon: EzIconTypes;
+  halfIcon: EzIconTypes;
+  fullIcon: EzIconTypes;
```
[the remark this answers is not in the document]
danidewitt: I'm still thinking about this default icon bit, maybe we can just import the three free default icons and it be ok... I'll add it and we can test the performance of it.

spdonegan1 (approved):

src/components/EzRating/Implementations/EzRatingCustom/EzRatingCustom.tsx
danidewitt: we're doing what [mui does for rating](https://mui.com/components/rating/#accessibility) - they use a `span` with `role="img"` and `aria-label=`${label}``

noranda (changes requested):
This looks great!

I like the idea of having a default icon (star?). Maybe we can use a free-license icon for this so we don't run into issues. We also might want to think about future uses of this, for example, we might want to only round to the nearest whole icon, or we might want to make these interactive at some point. For now though, I think this works for the required use case.

_Requested changes:_
I would suggest adding some logic tests to make sure we are rounding the values correctly. The Playroom is also not working correctly as it cannot import the icons.

maggiedbaker (approved):
looks good! I also like the idea of defaulting to a star icon, but either way is fine. Just curious, do we have any use cases currently that wouldn't be a star?

src/components/EzRating/EzRating.types.ts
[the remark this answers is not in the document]
danidewitt: @noranda and @maggiedbaker i think this thread answers your questions

danidewitt:
@noranda and @maggiedbaker  most of your questions are answered in the [thread](https://github.com/ezcater/recipe/pull/763#discussion_r814007308) above.

- adding default icon: this is definitely a performance issue, but also potentially a licensing issue.
- cases that aren't a star: yes, we have price ratings implemented several different ways in other projects, and i can imagine having heart rating or other icon ratings
- rounding to a whole number: if you want to round to something less precise than 0.5, i think this is fine to happen when you use the icon, so `Math.floor(value)` for your `value` prop (or `Math.ceil()`, you can decide if you want to be less precise)
- making it interactive: we can always add another interactive implementation of ratings later (maybe it's even a separate component that uses this read-only component)

@noranda i added tests and fixed the playroom

danidewitt:
visual tests passing
Screen Shot 2022-02-24 at 4 29 48 PM

noranda (changes requested):
The logic tests and playroom look great!

One last suggestion:

We should probably have boundaries for the values that extend outside the `0 - max` range. Currently a value of 100 will render 100 stars. We could restrict the values just in case (if value > max, value = max & if value < 0, value = 0).

danidewitt:
@noranda good catch - i added a check for value > max and value < 0 and tests for both of those cases.

danidewitt:
visual tests passing 
Screen Shot 2022-02-25 at 11 16 34 AM

noranda (approved):
Good to go!

danidewitt:
tests passing
Screen Shot 2022-02-25 at 6 17 52 PM

PR #767, opened 2022-03-01
docs: pin ghpages npm version, update related lock files and gatsby, update changelog

danidewitt:
## What did we change?
- Revert addition of the `.tool-versions` file https://github.com/ezcater/recipe/pull/737
- Updated related lock files and gatsby version related to the wrong npm version
- Updated changelog

## Why are we doing this?
- Doc-site deploys were failing

### Before

Screen Shot 2022-02-28 at 8 52 58 PM

### After
Screen Shot 2022-02-28 at 8 50 35 PM

## Checklist


morgannegagne (approved):

pgrippi (approved):

evan-007 (approved):
😭 sorry! is there another way to enforce a node version for recipe? it was tricky to figure out the right setup for local dev

pgrippi:
I wonder if a better fix than removing tool-versions is to just enforce the use of npm@6 (which is the last good version before they changed how peer deps works). If you were on node 16 you can use `corepack` for enforcing a specific version of npm/yarn, but since you're on 14 the best way is to probably add a build step inbetween these two: https://github.com/ezcater/recipe/blob/17ce7244a23e69dba15f89f70142094c85aac8eb/.github/workflows/ghpages.yml#L21-L27

new command:

```yaml
      - name: Install NPM
        run: npm i -g npm@^6
```

danidewitt:
@evan-007 good news i brought the .tool-versions back because @pgrippi suggested pinning npm in the ghpages action and that worked. thank you both! 

and @morgannegagne for a review too!

pgrippi (approved):
It might be good to add the same `npm i` command to the other github actions, it looks like they all run `npm ci` and will probably run into the same problem eventually

lgtm!

PR #774, opened 2022-03-04
fix: removed "require" function, a webpack protected word which was preventing loading of the doc site

noranda:
## What did we change?

This PR moves the `require` function directly into the scope that uses it so as not to redefine the protected `require` word.

## Why are we doing this?

Recipe now uses node version `14.18.3`, which has various key/protected words like `require`. We were redefining `require` in the doc site, creating an error in development and preventing the site from loading. 

## Checklist

### DIFF
--- a/.changeset/fuzzy-feet-shop.md
+++ b/.changeset/fuzzy-feet-shop.md
@@ -1,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+fix: removed "require" function, a protected word in node 14 which was preventing loading of the doc site
--- a/.changeset/fuzzy-feet-shop.md
+++ b/.changeset/fuzzy-feet-shop.md
@@ -2,4 +2,4 @@
 '@ezcater/recipe': patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/fuzzy-feet-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+fix: removed "require" function, a protected word in node 14 which was preventing loading of the doc site
```
pgrippi, line 5: I think this word is protected in webpack, not node - though I honestly don't know why this started breaking out of the blue

pgrippi (approved):

danidewitt (approved):
aside from @pgrippi 's comment about the changeset, please 🚢 !

danidewitt (approved):
🚀 🚀 🚀

pgrippi (approved):

.changeset/fuzzy-feet-shop.md
noranda: Whoops. Fixed. :)

PR #781, opened 2022-03-10
fix: options and children types for EzChoice

danidewitt:
## What did we change?
- Updated type definitions for `EzChoice` (if using `options`, don't use `children`, and if using `children`, don't use `options`)

## Why are we doing this?
- Downstream application failures 

## Screenshot(s) / Gif(s):
- N/A (private builds)

Passing visual tests:
Screen Shot 2022-03-09 at 10 43 53 PM

## Checklist


pgrippi (approved):

stacy-wells (approved):
Thanks for such a quick fix!

PR #786, opened 2022-03-15
fix: added 0 margin to text inputs to fix safari bug

noranda:
## What did we change?

Added a default `0` margin to `EzField` inputs.

## Why are we doing this?

Fixes Issue #712 which describes a bug only found in Safari where a `2px` margin is added to text inputs to make room for focus outlines. This PR overrides that margin to keep inputs aligned correctly. 

## Screenshot(s) / Gif(s):

Before:
Screen Shot 2022-03-15 at 11 19 27 AM

After:
Screen Shot 2022-03-15 at 11 19 36 AM

## Checklist


BoboFraggins (dismissed):
👍

pgrippi (dismissed):

pgrippi (approved):

danidewitt (approved):

.changeset/shiny-ears-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+fix: added margin to text inputs to fix safari bug
```
danidewitt, line 5: on the fence if this is `fix` or `style` - good to go either way!

noranda: Good point! I think a `fix` makes sense here because it fixes a bug, but it's definitely style-related. I'll leave it just because it's easier.

PR #799, opened 2022-04-05
Fix undefined exports for theme types

pgrippi:
## What did we change?

This PR fixes a bug where exported TypeScript types were making their way into the final JS build output, resulting in errors like this when building apps that consume recipe:

```
./node_modules/@ezcater/recipe/dist/esm/themes/index.js
export 'EzPalette' (reexported as 'EzPalette') was not found in './ezColors' (possible exports: ezColors, ezPalette)
```

The fix was just to use `export type` when exporting/reexporting the types, but prettier 1.0 doesn't support this syntax (it is a typescript 3.8 feature, and prettier 2.0 is the first version that adds support for it)

I also noticed that `babel` was outputting all of the unit tests in the final build, which seems less than ideal so I've updated the babel config so as not to waste time building tests!

## Why are we doing this?

This fixes build errors discovered when testing the recipe 14 canary build

## Checklist


.babelrc
```diff
@@ -15,5 +15,6 @@
       ]
     }
   },
-  "comments": false
+  "comments": false,
+  "ignore": ["**/__tests__"]
```
pgrippi, line 19: This is the change to babel that prevents anything in the `__tests__` folders from making their way into the final build - it doesn't affect jest, just the final output that the consumers of recipe use

.storybook/main.js
```diff
@@ -1,12 +1,9 @@
 const path = require('path');
 
 module.exports = {
-  "stories": [
-    "../src/**/*.stories.mdx",
-    "../src/**/*.stories.@(js|jsx|ts|tsx)"
-  ],
-  "addons": [
-    "@storybook/addon-links",
+  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
```
pgrippi, line 4: After upgrading prettier I did an `npx prettier -w .`, which runs prettier on all files in the repo, applying new formats that come in the 2.x version of prettier. The only things changed in the majority of these files are prettier formatting the code, there should be no functional changes

src/components/FocusScope/FocusScope.tsx
```diff
@@ -300,6 +300,8 @@ function getFocusableTreeWalker(root: HTMLElement) {
         return NodeFilter.FILTER_SKIP;
       },
     },
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
pgrippi, line 304: This is something I'd like to highlight - this is the same fix we made in the store-next copy of FocusScope. `createTreeWalker` no longer accepts a fourth parameter [(MDN shows it as deprecated)](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker#expandentityreferences), but for IE11 support it is needed. I've opted to just tell typescript to ignore this as it doesn't hurt evergreen browsers, I'll let ya'll decide when to actually remove support for it

noranda: This makes sense to me. Even if we don't officially support IE11, it doesn't hurt to support it here.

pgrippi:
I upgraded store-next to use this branch's [canary version](https://github.com/ezcater/recipe/releases/tag/v0.0.0-canary-20223514453) and confirmed that:

* The typescript types all still work with the new themes
* The build error around the `EzPalette` export has been fixed

store-next PR for reference: https://github.com/ezcater/store-next/pull/2225

danidewitt (approved):

PR #809, opened 2022-04-08
refactor: remove stitches package and import directly from npm

noranda:
## What did we change?

This PR removes the `@stitches` package from Recipe and instead imports it directly from npm. 

## Why are we doing this?

Recipe had previously added the `@stitches` library as a package in order to extend it with a polyfill to support IE11. As Recipe no longer supports IE11 and to make future work to reorganize its packages into workspaces easier, the package was removed and added as a dependency instead.

Note: `@stitches` was not upgraded as part of this due to significant effort that would be needed to do so and Recipe's plans to slowly phase out its use.

## Checklist


package.json
```diff
@@ -92,7 +92,6 @@
     "build:esm": "cross-env BABEL_ENV=esm babel src --out-dir dist/esm --extensions \".ts,.tsx\" --source-maps inline --copy-files --ignore '**/__tests__'",
     "build:watch": "npm run build:cjs && npm run build:esm -- --watch",
     "build:docs:prod": "cd doc-site && npm run build:prod",
-    "copy:packages": "cd packages/@stitches/core && npm run package",
```
pgrippi, line 95: you'll need to remove this command from the other commands - for instance `build` above runs:

```
npm run clean && npm run copy:packages && ...
```

pgrippi (dismissed):

pgrippi (approved):

package.json
```diff
@@ -85,14 +85,13 @@
     "start:with-ie11": "concurrently \"npm run build:watch\" \"(cd doc-site/ && npm run develop:with-ie11)\" --kill-others",
     "type-check": "tsc --noEmit",
     "type-check:watch": "npm run type-check -- --watch",
-    "build": "npm run clean && npm run copy:packages && npm run build:types && npm run build:cjs && npm run build:esm && cp .npmignore dist",
+    "build": "npm run clean && npm run build:types && npm run build:cjs && npm run build:esm && cp .npmignore dist",
```
danidewitt, line 88: this is probably right, just double checking though, are we sure we don't have to copy any other packages?

danidewitt: oh, nevermind! i see peter's resolved comment. proceed 🎉

PR #811, opened 2022-04-17
Yarn Workspaces

noranda:
## What did we change?

This PR does the following:
- Moves the Recipe monorepo to `yarn workspaces` with `lerna`, reorganizing as follows:
```
root
  docs/
  packages/
    analytics-bookmark/
    babel-plugin-recipe/
    eslint-plugin-recipe/
    recipe/
    recipe-ui-legacy-theme/
    recipe-ui-theme-marketplace/
```
  - As part of this reorganization and move to `yarn`, the following changes were also made:
    - Updates all relevant documentation to change `npm` scripts to `yarn`.
    - Updates github workflows to use the new `yarn workspaces` scripts.
- Removes the `css-module-types` package which was cruft from when recipe used `css` files.
- Creates a new root `package.json` file to handle top level commands such as:
  - `yarn develop`: runs all `develop` scripts in all packages, including building and running the doc site
  - `yarn test`: runs all `test` scripts in all packages
  - `yarn lint`: runs lint on the entire monorepo
- As a result of moving to `yarn workspaces`, existing dependency conflicts were corrected. This uncovered some issues, primarily due to a mismatch of dependency versions between packages, a result of incorrectly managing dependencies in a monorepo. This PR corrects these issues, including:
  - Changesets were only being applied to the root `recipe` package. Changes to other packages were not correctly applied including version bumps. This has been fixed. Running `yarn changeset` in the root will now prompt selection of affected packages and bump them appropriately.
  - A mismatch of `emotion` versions was causing issues in the doc site even though we upgraded to v11, due to dependency conflicts with `gatsby-plugin-emotion` and missing `@emotion/server`. This has been fixed by upgrading the plugin, adding the missing dependency, and code changes to bring in line with v11.
- Removes IE11 polyfills and functions which are no longer supported.
- Upgrades the doc site's `react-dom` dependency to match `react` (these versions should always match).
- Fixes a bug in the doc site `Layout` component when selecting active themes (`localStorage` values were conflicting between the server and client sides, crashing the doc site).
- Updates the `Lato` google font in the doc site to the latest version.
- Creates a separate `Placeholder` component in the doc site. Previously we were reaching into recipe to import, however this is not an exposed component and so was creating issues with yarn workspaces. We already do this for similar components like `Sprites`.
- Adds `resolutions` key to the root `package.json` file to keep typescript dependencies in sync between packages.
- Moves `playroom` into its own package.
- Moves `recipe-ui-legacy-theme` and `recipe-ui-theme-marketplace` packages to use `tsup` instead of `babel` because `parcel` was removed. `tsup` is recommended by ROG over `babel`. Note: Recipe can move to `tsup` in the future, but requires more work due to some issues with tree shaking and so was left alone.
- Updates `jest` to fix issues that arose after dependency conflicts were resolved.
  - Moves `jest axe` util into test util to import directly instead of configuring in every test.
  - Removes/skips some `EzCarousel` tests that are currently broken, likely due to the `jest` upgrade. As `EzCarousel` is due to be refactored, we can revisit these tests at that time.
- Fixes tests, lint, and typescript errors.
- Splits up recipe's `tsconfig` into build and develop files so we don't compile unnecessary files.

## Why are we doing this?

Our current monorepo approach is cumbersome and creates issues with shared dependencies between projects like Recipe and the doc site. While monorepos typically make managing dependencies easier, we haven't been managing them correctly, often creating dependency conflicts that cause unpredictable and undesired behaviors. `yarn workspaces` with `lerna` addresses these issues.

- [Yarn Workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/) is used to optimize and link different packages together.
  - Creates a single `yarn.lock` file in the root directory, and uses it to manage all dependencies between multiple packages, instead of creating one for each.
  - All project dependencies are installed in one go. `yarn install` will install the dependencies of the root folder and then for every package.
  - To run a specific command in a selected workspace, you can run `yarn workspace <package-name> <command>`.
- [Lerna](https://lerna.js.org/), whose foundation is `yarn`, adds additional monorepo capabilities and is used to optimize its management including running test scripts and building packages.

Improvements:
- The Recipe monorepo will now correctly handle dependencies across packages, keeping track of them in a single root level lock file.
- All packages will now keep track of changesets and bump versions appropriately.
- Removes cruft including code for IE support, unused packages, and code now included in upgraded dependencies.
- Dependencies can now be installed from a single place by running `yarn install` in the root.
- Tests run faster after upgrading `jest`.
- Bug fixes.

## Testing

To test: run the following after checking out the branch in the root:
```bash
> git clean -dfx # WARNING: will delete all non-tracked git files, so commit or stash any changes first
> yarn install
```

Try the following commands:
```bash
> yarn test # will run all tests in all packages (currently only recipe has tests)
> yarn build # need to build before linting
> yarn lint # will run lint across the monorepo
> yarn develop # will build and run the doc site including the playroom
> yarn changeset # will run the changesets CLI, asking to select one or more packages
```

## Checklist


docs/gatsby-config.js
```diff
@@ -60,15 +60,7 @@ module.exports = {
       // polyfill.io
       resolve: 'gatsby-plugin-polyfill-io',
       options: {
-        features: [
-          'es2015',
```
danidewitt, line 64: ❓ we only kept one of these - is that because we don't support IE11 anymore?

pgrippi: Yup, I think this came from my suggestion to Noranda - she was asking about the polyfills and I mentioned that when we dropped IE11 support on the store-next side we only needed the `smoothscroll` polyfill (I think it's not implemented in Safari only), all of the others are not necessary for our current browser support matrix

danidewitt: sweet! i made this comment before i made it through all 1001 files. looks so good!

danidewitt:
in local testing, i'm running into:

1. running yarn test takes some of the tests over 2 minutes each, and the whole thing eventually ends up crashing with exit code 1. 
2. yarn develop has errors, `SyntaxError: The requested module 'tsup' does not provide an export named 'defineConfig'` with `@recipe-ui/theme-marketplace`
3. yarn lint hangs

did i miss a step (or more) besides `git clean dfx` and `yarn install` in the root?

noranda:
> did i miss a step (or more) besides `git clean dfx` and `yarn install` in the root?

Oops it should be `git clean -dfx`, sorry!

danidewitt:
@noranda ok yeah, that `-` really did help. all the tests pass, and i only see 1 lint error (`error  Unable to resolve path to module '@ezcater/recipe'`).

i am still seeing an issue with `yarn develop` and `yarn build`: `SyntaxError: The requested module 'tsup' does not provide an export named 'defineConfig'`

danidewitt:
@noranda ok i can confirm it works **in this order** now:

```
> git clean -dfx
> yarn install
> yarn build
> yarn lint
> yarn develop
```

danidewitt (approved):

pgrippi (approved):
🚀

PR #812, opened 2022-04-19
fix: correct version on recipe docs homepage

noranda:
## What did we change?
Updates the latest recipe version on the doc site to be the latest version of recipe, not the monorepo.

## Why are we doing this?
With the latest change to `yarn workspaces`, a new top level `package.json` file was created to handle root scripts, but its version is not the same as recipe, which was moved to its own package in `packages/recipe`. This fixes the version pulled and displayed on the doc site to match recipe's version.

## Screenshot(s) / Gif(s):

Before:
Screen Shot 2022-04-19 at 4 44 00 PM

After:
Screen Shot 2022-04-19 at 4 44 23 PM

## Checklist


danidewitt (approved):
👍

pgrippi (approved):

PR #815, opened 2022-04-20
fix: deploy workflow will no longer run on PR merge

noranda:
## What did we change?
Updates the `deploy` github workflow to never run on PR merges.

## Why are we doing this?
We currently have the following release-related github workflows:
1. `canary` - creates a canary release on any PR merged into `main` except for Version Packages PRs
2. `deploy` - creates a canary release  on any PR that adds a `deploy` label
3. `release` - creates a normal release on Version Packages PRs merged into `main`

We ran into an issue recently where we merged a Version Packages PR with a `deploy` label which tried to run both the `deploy` and `release` workflows above. This created a conflict where the `deploy` workflow ran and [attempted to release a canary tag](https://github.com/ezcater/recipe/runs/5823101176?check_suite_focus=true#step:7:54), but instead [published the normal version](https://github.com/ezcater/recipe/runs/5823101176?check_suite_focus=true#step:7:58). It then ran the `release` workflow and wasn't able to publish to npm correctly as it had already been published.

The simple fix for this is to prevent the `deploy` workflow from running on any PR merge. Since the `canary` workflow will create a canary release for any PR merged into `main` that isn't a Version Packages PR, and we never need to create a canary release on a Version Packages PR, we only need the `deploy` workflow to run on PRs that haven't been merged by adding the `deploy` tag (in cases where we want to test before merging).

EDIT: @tristandunn pointed out that just blocking the `deploy` workflow from running on merged PRs will not solve the issue as the problem is tied to running the `deploy` workflow in Version Packages PRs specifically. The fix is therefore to prevent the workflow from running at all in a Version Packages PR, which shouldn't be needed anyways as the resulting canary version would be the same as the most recent canary version. Updated to fix.

## Testing
To test, we should keep an eye on the next few PRs and make sure the workflows run as expected. This should only affect when the `deploy` workflow is or is not run.

I've also added a `deploy` label to this PR. Once added, the `deploy` workflow ran as expected. When it's merged, if working correctly, it should not run the `deploy` workflow again.


pgrippi (dismissed):

noranda:
> ### ⚠️ No Changeset found
No changeset needed as this doesn't change any packages.

tristandunn:
Getting to read this off mobile now and I'm not 100% sure this fixes it.

The canary release that published a proper release wasn't due to merging, right? It was due to being label as deploy, triggering the deploy workflow, and resulting in a proper release. And I think that occurs because the release PR has no changesets, so it think it's time to release:

> No changesets found, attempting to publish any unpublished packages to npm

Which I think means you can't run the deploy task for a release PR without generating a proper release.

noranda:
> Which I think means you can't run the deploy task for a release PR without generating a proper release.

@tristandunn I think you're right! In this case, I think the solution is to prevent the `deploy` workflow from ever running in a Version Packages PR. Since the `canary` workflow will always be run whenever a PR (which isn't a version packages PR) is merged into `main`, we shouldn't need to create a canary version before releasing. We can just use the last canary release to test.

pgrippi (approved):

danidewitt (approved):

PR #818, opened 2022-04-21
refactor: remove version headers in recipe changelog

noranda:
## What did we change?
Removes major version headers from Recipe's changelog.

## Why are we doing this?
Recipe's changelog is autogenerated from our `changesets` library as part of our release process, however this does not include major version headers to separate the major versions. The `changesets` library provides functions to custom format the changelog messages, but does not provide any way to add these headers. Until now, we have been adding them manually, which we have determined is an unnecessary step to remember to do after every major release. This PR removes all the existing major version headers, making this an entirely automated changelog.

## Screenshot(s) / Gif(s):
Before:
Screen Shot 2022-04-21 at 1 13 40 PM

After:
Screen Shot 2022-04-21 at 1 13 56 PM

## Checklist


pgrippi (approved):

packages/recipe/CHANGELOG.md
```diff
@@ -656,10 +647,6 @@
 - created private css-module-types package to allow provide TypeScript definitions for CSS imports.
 - added top-level `repository` and `types` fields to Recipe's package.json.
 
----
-
-## Version 10
-
```
danidewitt, line 662: did we never have version 9 header? haha

danidewitt (dismissed):

packages/recipe/CHANGELOG.md
noranda: Actually, for some reason we don't even have any v9 releases! We skip from `8.0.3` right to `10.0.0`. We are even missing the "Migrate to Version 9" guide. Not sure what happened there, any ideas? @danidewitt

danidewitt (approved):

PR #819, opened 2022-04-27
Fixed build script to run synchronously

noranda:
## What did we change?
This PR adds `--concurrency=1` to the monorepo's build script.

## Why are we doing this?
The `recipe-ui-theme-marketplace` package has `recipe` as a dependency and therefore `recipe` needs to build first. Our `canary` build is currently failing because the two packages are building asynchronously.

No changeset needed.


pgrippi (approved):

danidewitt:
could be a [workflow](https://github.com/ezcater/recipe/blob/main/docs/src/pages/guides/versioning-and-changelog.md) changeset 🤷‍♀️ LGTM either way

danidewitt (approved):

noranda:
> could be a [workflow](https://github.com/ezcater/recipe/blob/main/docs/src/pages/guides/versioning-and-changelog.md) changeset 🤷‍♀️ LGTM either way

Maybe? It isn't a change in any of the packages though, so I wouldn't know which to attach it to.

PR #820, opened 2022-04-28
Replace Lerna with Turborepo

noranda:
## What did we change?

This PR replaces [Lerna](https://lerna.js.org/) with [Turborepo](https://turborepo.org/).

## Why are we doing this?

The [build command](https://github.com/ezcater/recipe/runs/6199170966?check_suite_focus=true) used by Github Actions has started failing in a weird way, where dependencies are being ignored and the packages are being built in the wrong order.

Last week, Lerna updated their [readme](https://github.com/lerna/lerna/commit/8b99493ac90484e05f7ebee533c0763534d60218) to say that the project is no longer being actively maintained. Since we only leverage Lerna for coordinating command execution orders (we use `changesets` for publishing and changelogs), we can use Turborepo instead, which only handles coordination, making it a slimmer alternative. Turborepo also comes with a `--graph` feature that outputs a tree graph showing the order in which commands will be executed. It's also a [Vercel product](https://vercel.com/blog/vercel-acquires-turborepo), and so is expected to have broad support.

## Screenshot(s) / Gif(s):

The GraphViz diagrams below verify the command execution orders are being handled correctly with Turborepo.

`yarn build --graph`

graphviz (1)

`yarn type-check --graph`

graphviz (2)

`yarn build '--filter=!@ezcater/recipe-analytics-bookmarklet' '--filter=!@ezcater/recipe-doc-site' '--filter=!@ezcater/recipe-playroom' --graph` (used by Github Actions)

graphviz (3)

## Checklist


noranda:
Tested this by adding a `deploy` label to this PR and watched the build, which executed in the correct order. No canary release created due to having no changeset as this isn't updating any packages.

danidewitt (changes requested):
i think we should have a `workflow` [changeset](https://github.com/ezcater/recipe/blob/main/docs/src/pages/guides/versioning-and-changelog.md)

pgrippi (approved):

danidewitt (approved):

PR #824, opened 2022-05-11
Sync package versions when doing a `changeset` version and release

pgrippi:
## What did we change?

This adds a new script, `bin/sync-package-versions`, which will manually update the package versions in the various package.json files so that they match whatever new version is created by the `changeset version` command

## Why are we doing this?

Yarn workspaces require that the package versions be kept in sync with each other. There is [a known bug in the changesets library](https://github.com/changesets/changesets/issues/432) when using yarn workspaces, where changesets doesn't update the various versions.

This has impacted you with both lerna and turborepo, where the release [fails because the packages are built out of order](https://github.com/ezcater/recipe/runs/6389531789?check_suite_focus=true).

The reason they are being built out of order is because the package versions no longer match after running `changeset version`, and these libraries are no longer able to see the interdependencies and thus run the scripts out of order.

This is what a typical build graph looks like (`yarn build --graph`):

before-canary

On main, this is what the graph looks like after running `yarn changeset version --snapshot canary && yarn build --graph`:

after-canary

As you can see, all packages are now seen as top-level, meaning that their build scripts can all run in parallel - which is wrong. After running `bin/sync-package-versions` the graph goes back to the original, where packages like the doc site and the marketplace theme will first wait for recipe to build

I don't love making a custom script to do this for us, but given that the changesets library doesn't support yarn workspaces I don't see a way around it without either moving off of yarn or moving to a different changeset management tool. For now this will fix your github actions and will let you run releases / canary releases without issues.

## Checklist


noranda (approved):
I think this is a good fix for the builds until we can look further into whether we might want to be using a different changeset library. This bug is nearly 2 years old. All tests pass.

danidewitt (approved):

PR #825, opened 2022-05-12
workflow: fix release script so that it doesn't release before merging the version changes PR

pgrippi:
## What did we change?

This removes the `yarn changeset version` and `bin/sync-package-versions` commands from the release script, and bumps the version of recipe to 12.0.5 so that the next release will be 12.0.6

## Why are we doing this?

When I updated the release script in #824 I did not realize calling `yarn changeset version` would break it. The github changeset action has magic built into it that determines it if should update the version PR or instead release to npm/github. It does this by checking for any version changes, which the newly introduced `yarn changeset version` does

Sadly when I merged the above PR it released 12.0.4 (`main`) as 12.0.5 but without any of the package version bumps, which will result in errors if consumers were to upgrade. I have since pulled that version, and as we've learned we cannot re-release a version so the only option is to skip it

I am really sorry I broke the release script!

## Checklist


pgrippi:
Reading the [`changeset` docs on publish](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish):

> Because this command assumes that last commit is the release commit you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the releases changes are acurate.

I believe the way the Version Packages PR is made is changesets:

* Runs `yarn changeset version`
* Commits the change to the `changeset-release` branch
* Pushes it up

So once that PR is merged, `yarn changeset version` will do nothing because the changes are already applied and there are no more changesets in `.changesets`, and because there is nothing changed it knows to publish the release

By us running that command we've successfully tricked it into thinking we merged the Version Changes PR without actually doing that 😩

noranda (approved):

danidewitt (approved):

PR #827, opened 2022-05-13
Remove snitches

noranda:
_Note: This PR should be merged into the feature branch `v15`._

## What did we change?
This PR removes `snitches` from Recipe, resulting in a breaking change.
- Adds a migration guide for v15.
- Updates the Getting Started page in the doc site with instructions for adding styles for apps with SSR.
- Removes `snitches` from all Recipe components and tests.
- Upgrades `stitches` and reconfigures Recipe to use without `snitches`.
- Disables by default all visual snapshots, which broke as a result of upgrading `stitches`. Fixing will either require a significant effort to upgrade [sosia](https://github.com/ezcater/sosia), or by moving to [Chromatic](https://www.chromatic.com/), which is planned and should occur as part of the v15 release.
- Fixes broken images in `EzAppLayout` doc page.

## Why are we doing this?
With the decision in [Version 14](https://ezcater.github.io/recipe/support/migrating-to-recipe-14) to move away from `stitches`, `snitches` would have eventually been removed with it, however it turns out `snitches` is incompatible with React 18 which adds [stricter hydration errors](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#other-breaking-changes), preventing both Recipe and downstream apps from upgrading.

Removing `snitches` makes upgrading to React 18 possible, however it does require some additional configuration for apps using server-side rendering (SSR). If a downstream app uses SSR, extracted styles from `stitches` will now need to be inserted directly into the html of the app.

For example, if using Next.js, you might add the styles as follows:

  ```js
  import React from 'react';
  import NextDocument, {Html, Head, Main, NextScript} from 'next/document';
  import {stitches} from '@ezcater/recipe';

  export default class Document extends NextDocument {
    render() {
      return (
        <Html lang="en">
          <Head>
            <style id="stitches" dangerouslySetInnerHTML={{__html: stitches.getCssText()}} />
          </Head>
          <body>
            <Main />
            <NextScript />
          </body>
        </Html>
      );
    }
}
```

## Testing
Verify that all tests pass and all components look and behave the same. Visual snapshots can be reviewed in [Chromatic](https://www.chromatic.com) once set up. We will want to test this with a few downstream apps as well to make sure the SSR config works as expected. 

Given the number of changes in this PR, I have broken it up into separate commits to aid with review.

## Checklist


danidewitt (commented):
nice! 

i think we'll have to update the following (not because of any changes currently, but other things being merged):

1. after the carousel is merged: update the migration doc to include the breaking changes from the carousel changes
2. i assume this has been tested downstream in next and react apps?
3. after the carousel is merged and rebasing this PR: stitches-related changes from my 3 PRs (cards, buttons, carousel) 
4. any additional stitches-related changes from merging any of your other PRs before this one (with a rebase, obviously)
5. re-running tests to make sure it doesn't introduce any new changes from other PRs merged before this one

pgrippi:
> i assume this has been tested downstream in next and react apps?

yup, we tested it in store-next and store-react, both were able to generate the inline stitches css like the docs recommended. I didn't try ezmanage as I don't have it running locally, but I would expect it would work the same

noranda:
@danidewitt 

> * after the carousel is merged: update the migration doc to include the breaking changes from the carousel changes

Yes! We should do this as part of the carousel PR once this PR is merged into `v15`

> * i assume this has been tested downstream in next and react apps?

Yes! @pgrippi mentioned he tested this and all looks great!

> * after the carousel is merged and rebasing this PR: stitches-related changes from my 3 PRs (cards, buttons, carousel)

Cards and buttons stitches changes were made in the latest commit for this PR. Carousel changes can happen in that PR after this is merged and the carousel PR is rebased.

> * any additional stitches-related changes from merging any of your other PRs before this one (with a rebase, obviously)

Done!

> * re-running tests to make sure it doesn't introduce any new changes from other PRs merged before this one

Done!

danidewitt (approved):

craigcav-stripe:
👋  Hey everyone - hope y'all doing well! 

I was just being a fly-on-the-wall, and noticed this PR. One of the goals of snitches - aside from supporting config-free support for server rendering - was to support streamed server rendering. This becomes more important in React 18, where progressive hydration of components is possible, and will likely become essential if recipe is to ever be compatible with [React server components](https://nextjs.org/docs/advanced-features/react-18/server-components).

There are three major strategies in server rendering styles:

1. style extraction to css files (build time)
2. flush styles to an inline style element in the head (requires full server response before styles can be printed)
3. append styles in small style or script tags as you render (streaming)

Snitches attempts to do number 3 (but as you've called out would need updating to avoid hydration errors in React 18). This PR switches the approach to number 2, which ultimately will cause issues should anyone want to support streamed server rendering and progressive hydration. 

This issue might go-away once you're fully migrated back to emotion, but in the meantime I figured I'd call it out to put it on y'alls radar!

noranda:
Hi Craig, thanks for reaching out!

We are indeed aware that this solution doesn't fully support allowing users to use React 18's new streaming API. Ideally this would be something supported by `@stitches/react`, but it [doesn't seem like they have any plans to support it](https://github.com/modulz/stitches/issues/1026#issuecomment-1138900012). Our immediate goal here is to unblock users from upgrading to React 18. In the mean time, users will have to either extract the styles into the head as you've pointed out, or they can use the steaming API with the flash of unstyled content until React is downloaded and starts up.

Our plan is to incrementally remove stitches, converting all components to use emotion, primarily through MUI, which will allow us to work with React 18 streaming without the flash of unstyled content. If stitches adds native support for static extraction in the near future, then we might not need to prioritize this work, but it's been [talked about it for a long time](https://github.com/modulz/stitches/issues/863#issuecomment-977744554) with no visible progress towards implementing it.

From our initial testing of this branch in next.js, it appears that the `getCssText()` function returns the CSS for all calls of `theme.css()` from the top-level modules, which might also provide a path forward for users who want to try streaming. They can inject a `<style>` tag in the head tag as part of the initial response regardless of what components were rendered (or to-be rendered). The output of `getCssText()` seems to grow and shrink as components are tree shaken out of the final build - but again this is only an initial observation.

PR #832, opened 2022-05-19
draft feedback for carousel PR

danidewitt:
## What did we change?
⚠️ DO NOT MERGE ⚠️ 
The purpose of this PR is to try out suggested changes in https://github.com/ezcater/recipe/pull/831.

Specifically these comments:
- [x] [items vs children](https://github.com/ezcater/recipe/pull/831#discussion_r876399299)
- [x] [handlePageChange](https://github.com/ezcater/recipe/pull/831#discussion_r876415584)
- [x] [updateResponsiveStates](https://github.com/ezcater/recipe/pull/831#discussion_r876425580)

## Why are we doing this?
The purpose of this PR is to try out suggested changes in https://github.com/ezcater/recipe/pull/831.


packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -33,35 +35,37 @@ const EzCarouselCustom = ({items, title, link, description, onPageChange, ref})
...
+    },
+    [setPage, onPageChangeRef]
+  );
 
   useEffect(() => {
     if (page !== scrollPosition) {
       handlePageChange(page, scrollPosition);
     }
-  }, [scrollPosition]);
+  }, [page, scrollPosition, handlePageChange]);
```
danidewitt, line 59: this piece seems to introduce this [issue](https://github.com/ezcater/recipe/pull/831#discussion_r877473104), even though it's now more stable @pgrippi thoughts? (i understand WHY its happening)

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -24,7 +24,9 @@ import {
 } from './EzCarouselStyles';
 import {useUniqueId} from '../../../../utils/hooks';
 
-const EzCarouselCustom = ({items, title, link, description, onPageChange, ref}) => {
+const EzCarouselCustom = ({children, title, link, description, onPageChange, ref}) => {
+  const itemsRef = useRef(React.Children.toArray(children));
```
pgrippi, line 28: the value you pass to `useRef` is only used on the first call, you'll need to assign `itemsRef.current` below this like you do below. you can shuffle items to the top:

```jsx
const items = React.Children.toArray(children);
const itemsRef = useRef(items);
itemsRef.current = items;
```

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -24,7 +24,9 @@ import {
 } from './EzCarouselStyles';
 import {useUniqueId} from '../../../../utils/hooks';
 
-const EzCarouselCustom = ({items, title, link, description, onPageChange, ref}) => {
+const EzCarouselCustom = ({children, title, link, description, onPageChange, ref}) => {
```
pgrippi, line 27: this component is untyped, so your arguments are untyped

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -33,35 +35,37 @@ const EzCarouselCustom = ({items, title, link, description, onPageChange, ref})
...
+  );
 
   useEffect(() => {
     if (page !== scrollPosition) {
       handlePageChange(page, scrollPosition);
     }
-  }, [scrollPosition]);
+  }, [page, scrollPosition, handlePageChange]);
 
   const updateResponsiveStates = useCallback(() => {
```
pgrippi, line 61: I can't comment on the right line because it's outside the diff, but now that this function is stable (it's only dependency is a ref), the resize event listener can be updated so that it runs on only mount + dismount:

```jsx
useEffect(() => {
  window.addEventListener('resize', updateResponsiveStates);
  return () => {
    window.removeEventListener('resize', updateResponsiveStates);
  };
}, [updateResponsiveStates);
```

pgrippi:
this is looking good!

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
pgrippi: I _think_ you want to use `isDesktopSize` here (or more specifically, `!isDesktopSize` in your if condition). The desktop carousel doesn't scroll, only the mobile one, so you'll constantly get a scroll position of `0`

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -82,6 +86,10 @@ const EzCarouselCustom = ({items, title, link, description, onPageChange, ref})
     handlePageChange(page, page - 1);
   };
 
+  const itemsBySlidesPerPage = calculateItemsBySlidesPerPage(
+    items,
+    outerContainer.current?.clientWidth || DEFAULT_RESPONSIVE_VALUES.minContainerWidth
+  );
```
pgrippi, line 92: I think this is a place we could optimize as well. `calculateItemsBySlidesPerPage` chunks items into arrays like this:

```
children = [s1, s2, s3, s4, s5, s6];
itemsBySlidesPerPage = [[s1, s2, s3], [s4, s5, s6]];
```

which made sense when you were using it as a piece of state, but now you're creating nested arrays just to loop over a portion, and you're doing this every time you render.

you could instead use slice + page + the number of slides you want to show:

```jsx
const visibleItems =
  isDesktopSize
    ? items.slice(page * perPage, perPage).concat(...new Array(numberOfBlanks))
    : items;
```

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
danidewitt: omg this is such a simple solution! should have been obvious. i was trying to use `scrollRef.current.scrollLeft` logic... i even have the variable already 🤦‍♀️

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -82,7 +83,7 @@ const EzCarouselCustom = ({items, title, link, description, onPageChange, ref})
     handlePageChange(page, page - 1);
   };
 
-  const isDesktopSize = responsiveSize.includes('Desktop');
+  const itemsBySlidesPerPage = calculateItemsBySlidesPerPage(items, containerSize);
```
pgrippi, line 86: I think this is a place we could optimize as well. `calculateItemsBySlidesPerPage` chunks items into arrays like this:

```
children = [s1, s2, s3, s4, s5, s6];
itemsBySlidesPerPage = [[s1, s2, s3], [s4, s5, s6]];
```

which made sense when you were using it as a piece of state, but now you're creating nested arrays just to loop over a portion, and you're doing this every time you render.

you could instead use slice + page + the number of slides you want to show:

```jsx
const slidesPerPage = calculateSlidesPerPage(containerSize);
const visibleItems =
  isDesktopSize
    ? items.slice(page * slidesPerPage, slidesPerPage).concat(...new Array(numberOfBlanks))
    : items;
```

with the added benefit of skipping all of that array work on mobile!

you'll also have to change the logic for `hasMultiplePages` / `isFirstPage` / `isLastPage`, but you'll have page, slidesPerPage, and visibleItems.length which should give you everything you need to do those calculations!

danidewitt: i'll have to figure out the function for the desktop size - we only want blanks if it's the last page and there are empty spots that need to be added. 

right now this is showing 4 with 2 blanks on the first page and nothing on the second page (if you give it 6 slides and size the page for 4 per page)

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -24,53 +32,73 @@ import {
...
+  children,
+  title,
+  link,
+  description,
+  onPageChange,
+  ref,
+}) => {
+  const items = React.Children.toArray(children);
+  const itemsRef = useRef(items);
+  itemsRef.current = items;
```
pgrippi, line 55: after all this refactoring it turns out we don't need this ref! 😂

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -24,53 +32,73 @@ import {
...
+  const handleContainerResize = useCallback(() => {
+    setContainerSize(outerContainer?.current?.clientWidth);
+  }, [outerContainer]);
+
+  useEffect(() => {
+    handleContainerResize(); // the first time, the ref outerContainer is null so it shows 3 instead of 4 slides on a wide screen
+  }, [handleContainerResize]);
+
+  const responsiveSize = useMemo(() => calculateResponsiveSize(containerSize), [containerSize]);
+  const itemWidth = useMemo(() => responsiveValues[responsiveSize].itemWidth, [responsiveSize]);
```
pgrippi, line 74: looking at this now that it's been simplified I think this `useMemo` can go, `responsiveSize` is a string and you're just doing a straight object lookup here, that'll be way faster than a `useMemo` call:

```jsx
const itemWidth = responsiveValues[responsiveSize].itemWidth;
```

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
```diff
@@ -140,7 +170,7 @@ const EzCarouselCustom = ({items, title, link, description, onPageChange, ref})
                 sx={
                   isDesktopSize ? desktopListItemStyles(itemWidth) : mobileListItemStyles(itemWidth)
                 }
-                key={item && item.key ? item.key : index}
+                key={index}
```
pgrippi, line 173: 🎱 should this maybe be `index * (page + 1)`? I worry that when you change pages this `key` will always be 1 / 2 / 3 even though the item has changed

pgrippi (approved):

packages/recipe/src/components/EzCarousel/Implementations/EzCarouselCustom/EzCarouselCustom.tsx
danidewitt: this does seem to affect the issue where i quickly drag from small mobile to large desktop size and get that `map` error - so maybe we do need it? seems like we shouldn't

noranda (approved):
The changes here look good! I did notice a new bug though on this PR that I didn't see on the other one. If you horizontally scroll through the cards, once you get to the end, if you keep scrolling it will jump back to the beginning. Is this expected behavior?

EzCarousel-scroll

PR #839, opened 2022-06-13
adds form control element

ez-alexfrazer:
## What did we change?

- Adds EzFormControl, based on https://mui.com/material-ui/api/form-control/

## Checklist


packages/recipe/src/components/EzFormControl/Implementations/EzFormControlImplementation.tsx
```diff
@@ -0,0 +1,13 @@
+import React, {forwardRef} from 'react';
+import FormControl from '@mui/material/FormControl';
+import {EzFormControlProps, Ref} from '../EzFormControl.types';
+
+const EzFormControlImplementation = forwardRef<Ref, EzFormControlProps>((props, ref) => {
```
noranda, line 5: I would change this component name to `EzFormControlMui` to match our other implementations that use MUI.

packages/recipe/src/components/EzFormControl/EzFormControl.tsx
```diff
@@ -0,0 +1,19 @@
+import React, {forwardRef} from 'react';
+import {EzFormControlImplementation} from './Implementations';
+import {EzFormControlProps, Ref} from './EzFormControl.types';
+
+/**
+ * FormControl
+ * ===========
+ * Wraps elements that belong in a form
+ * @see https://github.com/ezcater/recipe/pull/839
```
noranda, line 9: I think this comment can be removed as its use will be documented and we don't need to refer to the PR in the code.

packages/recipe/src/components/EzFormControl/Implementations/EzFormControlImplementation.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return <FormControl {...props} ref={ref} />;
+});
+
+EzFormControlImplementation.defaultProps = {};
```
noranda, line 9: We can probably leave this line out if we don't actually have any default props.

packages/recipe/src/components/EzFormControl/__tests__/EzFormControl.test.tsx
```diff
@@ -0,0 +1,28 @@
+import React, {ComponentProps, FC} from 'react';
+import {axe, render} from '../../../../test-utils';
+import EzFormControl from '../index';
+
+const scope = {EzFormControl};
```
noranda, line 5: This line is no longer needed.

packages/recipe/src/components/EzFormControl/__tests__/EzFormControl.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+const Component: FC<Partial<ComponentProps<typeof EzFormControl>>> = props => (
+  <EzFormControl {...props} />
+);
+
+describe('EzFormControl logic', () => {
+  /**
+   * Insert your unit tests here to validate your component
+   */
+  it('should have tests', () => {});
+});
```
noranda, line 16: TODO: either add tests or remove

packages/recipe/src/components/EzFormControl/EzFormControl.md
```diff
@@ -0,0 +1,33 @@
+---
```
noranda, line 1: TODO: write the docs

packages/recipe/src/components/EzFormControl/EzFormControl.preview.md
```diff
@@ -0,0 +1,11 @@
...
+<div
+  style={{
+    '--zoom': 4,
+    '--shadow': 'none',
+    '--color': '#565a5c',
+  }}
+>
+  <EzFormControl />
+</div>
+```
```
noranda, line 11: I think we can probably remove this file. This is more of a util component that will be used with other components (like `EzRadio` and `EzCheckbox`). We will be redesigning how to show these types of components in the doc site in the future, but for now they won't need a preview.

noranda (changes requested):
Added some comments as an initial review while you work on this. Thanks!

noranda (changes requested):
Looking better! Just a few more small changes.

noranda (dismissed):
Looks good! Thanks!

packages/recipe/src/components/EzFormControl/EzFormControl.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {forwardRef} from 'react';
+import {EzFormControlMui} from './Implementations';
+import {EzFormControlProps, Ref} from './EzFormControl.types';
+
+const EzFormControl = forwardRef<Ref, EzFormControlProps>((props, ref) => {
```
danidewitt, line 5: do we want to be more explicit about some of the props we pass into this component? i'm thinking some of them should be, especially if we ever have other implementations (or if mui changes) @ez-alexfrazer @noranda

packages/recipe/src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export {default as EzCarousel} from './EzCarousel';
 export {default as EzCheckbox} from './EzCheckbox';
 export {default as EzField} from './EzField';
 export {default as EzFlashMessage} from './EzFlashMessage';
+export {default as EzFormControl, useFormControl} from './EzFormControl';
```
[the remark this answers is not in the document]
noranda: We don't want to expose this on the component grid because it's a util component, only used to support other components like `EzRadio`.

packages/recipe/src/components/EzFormControl/EzFormControl.tsx
noranda: Not a bad idea to keep things in scope. @ez-alexfrazer I would look into which props you need to support for `EzRadio` and only include those. You can see how I did the same thing in [`EzChip`](https://github.com/ezcater/recipe/blob/main/packages/recipe/src/components/EzChip/EzChip.tsx#L6) to limit what props get passed in.

danidewitt: which ones do you think we'll always need for any implementation of a form control (mui or not)? this ask is mostly so we don't create a mess of unknown downstream props that aren't explicitly named, which would created breaking changes and backwards compatibility issues in the future.

spdonegan1: my 2 cents. I like the idea of being explicit about the props. 

From the breaking change perspective my general mindset is just that it tends to be easier to add a prop than to remove a prop, so if we just start with a completely open set of props our hands are immediately tied. As an example for something like the `margin` prop on this one I could see a world where if recipe wants to in the future add some opinionated standard base margin it could then be in conflict with consumer's sending in their own margin prop. types would all pass but there would be potential for a style breaking change which is a lot harder to track down. 

To the point of upgrading mui, I would love a world where recipe could just update it's mui version without being a breaking change to consumers (and they may not even have to know it happened). These sorts of big lib major upgrades (mostly looking at emotion) have been a source of a lot of developer effort+some frustration to upgrade things and we have left plenty of repos on older versions of recipe because there isn't the time to deal with breaking changes so any way around not repeating this would be a huge win in my book.

I also like it from a documentation perspective in terms of being able to get a basic idea of what sort of props / controls I would have if i want to use this without needing to open the mui docs/ rely on my ide to show me the types. Just from looking at this code I wouldn't know how to use this component.

danidewitt: We don't need to upgrade anything here anytime soon. 

I'm specifically talking about the Recipe interface `EzFormControl`, not the MUI implementation `EzFormConrolMui`. 

We need specific props for our Recipe interfaces for a variety of reasons. 

An example: we have multiple implementations of some components - so we need a clear set of Recipe props for the interface. A good example of this is [EzIcon interface](https://github.com/ezcater/recipe/blob/main/packages/recipe/src/components/EzIcon/EzIcon.tsx), where we have [4 implementations](https://github.com/ezcater/recipe/tree/main/packages/recipe/src/components/EzIcon/Implementations) of it. 

I know we only *currently* have one implementation of this form control component now, but if we allow downstream apps to pass in any props, and we don't have an opinion about them, they're going to vary across apps, which is going to be a mess to sort out if we're introducing breaking changes at a later date. Examples: we could decide to be more opinionated  in the future, we decide to add another library's implementation, or we decide to rip out entirely MUI (which isn't unlikely given how quickly front ends change), or we decide we need a custom version because the designers want something very specific (or we want to switch css engines that support static css), or we start doing A/B testing or experimental components, or MUI introduces breaking changes, or some other implementation's library introduces breaking changes, then we're stuck with whatever is passed as `...props` which could potentially be a mess to sort out because we have 25+ repos using Recipe. 

Hopefully that all makes sense?

danidewitt: All those breaking changes possibilities aside, I also agree with @spdonegan1 about the documentation. We don't want devs or designers to need to be tracking down old MUI documentation or even current MUI documentation to know what props can and can't be used. This would require looking up Recipe's current version of a implementation's library (in this case MUI, but in other components, it could be font awesome or something else) and then tracking down that version of the library's documentation (if it's even available). This gets especially confusing with a component that has multiple implementations because the prop names are different across libraries (as you can see, [it's a real mess out there](https://open-ui.org/research/component-matrix)). So we need to define what we want the prop names to be for our own interfaces. Example, again, for `EzIcon`, props and prop names vary across our implementations of Font Awesome, ezCater icons, and MUI.

noranda: @ez-alexfrazer Let's keep this simple and start with `children` as the only prop for this PR. If we need to add props (like `error` or `color`) later as part of `EzRadio`, we can do that then. It's a lot easier to add a prop than remove one.

EzFormControl.jsx:
```js
...
const EzFormControl = forwardRef<Ref, EzFormControlProps>(({children}, ref) => (
  <EzFormControlMui ref={ref}>{children}</EzFormControlMui>
));
...
```

EzFormControlMui.tsx:
```js
...
const EzFormControlMui = forwardRef<Ref, EzFormControlProps>(({children, ...props}, ref) => {
  return (
    <FormControl {...props} ref={ref}>
      {children}
    </FormControl>
  );
});
...
```

EzFormControl.types.ts:
```js
...
export interface EzFormControlProps {
  children: ReactNode;
}
...
```

EzFormControl.test.tsx:
```js
...
const Component: FC<Partial<ComponentProps<typeof EzFormControl>>> = ({children}) => (
  <EzFormControl>{children}</EzFormControl>
);
...
```

noranda: Changes pushed.

noranda (approved):

packages/recipe/src/components/EzFormControl/EzFormControl.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {forwardRef} from 'react';
+import {EzFormControlMui} from './Implementations';
+import {EzFormControlProps, Ref} from './EzFormControl.types';
+
+const EzFormControl = forwardRef<Ref, EzFormControlProps>(({children}, ref) => (
```
danidewitt, line 5: ❓ do we minimally need `focused`, `erorr`, `disabled`, `required` here? what does this component do without those contexts?

danidewitt: another ❓ will we be able to use mui form control with implementations of child components that aren't mui?

noranda: @danidewitt

Q1: We may end up using those props when we implement `EzRadio` and/or `EzCheckbox`, but I think it's better to add those in when we've had time to decide exactly which are needed. Currently, the only required prop for this component is `children`. 

Q2: I'd have to look at the source code to see exactly how it's used, but if we needed a new child implementation that didn't use MUI, we'd could always create new implementations of these control components if we needed to.

danidewitt (approved):

PR #841, opened 2022-06-14
Restore Github Actions test suite

pgrippi:
## What did we change?

This PR does a few things:

* Adds a new Github Action workflow that will run the linting, type checking, and jest tests
* Adds a new shared Github Action for setting up node.js and yarn, with caching, and running `yarn install`, and refactors all existing workflows to use this shared action
* Upgrades jest to 28, which enables sharding of tests (splitting tests equally across different workers)
* Upgrades testing library packages to their latest versions supported by React 17
* Removes a few jest packages and config that were leftover from when we were running Recipe tests inside of our Jenkins environment and are no longer needed when running inside of Github Actions (`jest-junit`, `jest-standard-reporter`)

## Why are we doing this?

Now that the visual regression tests have been moved to Chromatic we can start running `jest` tests again in Github Actions without worrying about them randomly failing.

I opted to move all of the CI checks into a single action, which was [recommended to me by Jen and Alex](https://github.com/ezcater/store-next/pull/2335#discussion_r868290671) when I was toying around with using Github Actions in store-next.

## Checklist


packages/recipe/src/components/EzCarousel/__tests__/EzCarousel.test.tsx
```diff
@@ -82,9 +80,11 @@ describe('EzCarousel', () => {
   });
 
   // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('calls the onPageChange handlers when interacting with carousel pages', () => {
+  it.skip('calls the onPageChange handlers when interacting with carousel pages', async () => {
     jest.spyOn(console, 'log').mockImplementation();
 
+    const user = userEvent.setup();
```
pgrippi, line 86: You'll see this done across several test file (basically anything that used `userEvent`)

User event had a major 14 release, which deprecated calling `userEvent.xxx` directly, favoring instead calling `setup()` in each file and using that object in the tests. It also standardized on having every method (`click`, `type`, etc) return a promise, so every test using it needs to be `async` and every user event `await`ed

One last gotcha to this upgrade is that when a test uses `useFakeTimers` you have to call setup like: `setup({delay: null})`, otherwise the test will hang because user event internally uses `setTimeout`. Passing `{delay: null}` in bypasses the setTimeout call allowing everything to "just work"

packages/recipe/src/components/EzField/Open.tsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
-import userEvent from '@testing-library/user-event';
 
 const Open = ({children, containerRef}) => {
   React.useEffect(() => {
     const input = containerRef.current.querySelector('input');
-    userEvent.click(input);
+    input.click();
```
pgrippi, line 6: I opted to refactor this test component helper to _not_ use `userEvent`. This was causing the testing library to be pulled into the storybook code, and the new version 14 breaks the storybook build.

We can instead just use the native DOM `.click()` method to achieve the same behavior without a library!

packages/snapshot/repng.ts
```diff
@@ -31,7 +31,7 @@ const getHtmlData = (Component: React.ClassType<unknown, any, any>): string => {
     runScripts: 'dangerously',
   });
   const {document} = window;
-  global.window = window;
+  (global as any).window = window;
```
pgrippi, line 34: I don't know why this started failing the type-check. I did not upgrade typescript and I confirmed that typescript was not bumped, but typescript started complaining that `DOMWindow` cannot be assigned to `Window`.

This is a hacky way to get `window` to be set (setting it on `global`), so I think a hack to get typescript to allow it is also fine

.node-version
```diff
@@ -0,0 +1 @@
+14.18.3
```
noranda, line 1: Is there a reason we needed to add this when we already have the node version in `.tool-versions`?

noranda (approved):
This looks great! Thanks!

We should probably note that this removes Code Climate in the CI. I don't think we're really using Code Climate anyways I'm not sure how it would work with the test sharding splitting up the tests, so I think this is fine.

.node-version
pgrippi: Great question!

This stems from a limitation in the `setup-node` github action. If you look at the changed workflows you'll notice that they had this in them:

```yaml
- name: Setup Node.js 14.x
  uses: actions/setup-node@v3
  with:
    node-version: '14'
    cache: 'yarn'
```

The `node-version: '14'` will grab the latest node version (14.19.3), which is different than the node version you currently specify in your `.tool-versions` file. This isn't great, because it means your CI workflows are using a different node version than your local machine, which could lead to errors that you won't encounter locally

There are two ways to prevent this from happening, you can either set the specific version in the yaml file (`node-version: 14.18.3`), or you can tell the setup-node action to [read from a version file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file). The limitation here is that the setup-node action doesn't support the ASDF `.tool-versions` file, it only supports the NVM/nodenv dot files.

ASDF has a `legacy_version_file` config setting, which is off by default, which tells ASDF to read these individual dot files. If recipe told developers to set that config in their local `~/.asdfrc` file then we could remove the `.tool-versions` file and only use the `.node-version` file.

Instead of forcing a config option on developers I have, here and in other projects, just resorted to using multiple version files. This gives developers the freedom to use the tools they like, while also supporting the action config.

Upgrading node does get more complicated, as you now need to modify two files, but doing it this way at least ensures that the files you need to change are both in the same directory, not hidden deeply in the `.github` folder.

If you'd rather specify the version in the yaml file I am happy to make that change, or if you'd rather only have `.node-version` I would be fine with that too!

noranda: Makes sense to me! Thanks!

PR #842, opened 2022-06-14
Fix changeset output to filter out recipe deps and collapse commits

noranda:
## What did we change?
Fixes the changeset output to filter out recipe deps and collapse commits.

## Why are we doing this?
With the yarn workspaces rework, recipe was moved into its own package. As a result, any update in `@ezcater/recipe` now results in a dependency update in packages that use `@ezcater/recipe` (currently `@recipe-ui/theme-marketplace` and `@ezcater/recipe-snapshot`), which then gets logged in the changeset. These recipe dependency updates should not be logged as they don't reflect external dependency updates for the package, which is what we care about.

You can see this in the current [Version Packages PR](https://github.com/ezcater/recipe/pull/826) as a giant list of "Update dependencies" logs, one for each time `@ezcater/recipe` was updated:
Screen Shot 2022-06-14 at 2 14 57 PM

This PR filters out `@ezcater/recipe` dependencies from the changelog to correct this. To further help readability, it also combines dependency commits into a single line. In theory (hard to test), it should generate a changelog that looks more like this:

```
- Updated dependencies [sha1] [sha2] [sha3] [sha4]
```

## Checklist


djcp (approved):
Makes sense to me! Too much chatter in the changelogs

PR #843, opened 2022-06-16
basic EzFormLabel element

ez-alexfrazer:
## What did we change?

- Adds `EzFormLabel` component

## Why are we doing this?

- Compliance with mui

## Checklist


packages/recipe/src/components/EzFormLabel/EzFormLabel.tsx
```diff
@@ -0,0 +1,13 @@
+import React, {forwardRef} from 'react';
+import {EzFormLabelImplementation} from './Implementations';
+import {EzFormLabelProps, Ref} from './EzFormLabel.types';
+
+const EzFormLabel = forwardRef<Ref, EzFormLabelProps>((props, ref) => {
+  return <EzFormLabelImplementation ref={ref} {...props} />;
+});
+
+EzFormLabel.defaultProps = {};
```
noranda, line 9: Remove line.

packages/recipe/src/components/EzFormLabel/__tests__/EzFormLabel.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+const Component: FC<Partial<ComponentProps<typeof EzFormLabel>>> = props => (
+  <EzFormLabel {...props} />
+);
+
+describe('EzFormLabel logic', () => {
+  it('should have tests', () => {});
+});
+
+describe('EzFormLabel', () => {
+  visualSnapshots({markdown, scope});
```
noranda, line 18: Remove.

packages/recipe/src/components/EzFormLabel/__tests__/EzFormLabel.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React, {ComponentProps, FC} from 'react';
+import {visualSnapshots} from 'sosia';
+import {axe, render} from '../../../../test-utils';
+import markdown from '../EzFormLabel.md';
```
noranda, line 4: Remove.

packages/recipe/src/components/EzFormLabel/__tests__/EzFormLabel.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const scope = {EzFormLabel};
+
+const Component: FC<Partial<ComponentProps<typeof EzFormLabel>>> = props => (
+  <EzFormLabel {...props} />
+);
+
+describe('EzFormLabel logic', () => {
+  it('should have tests', () => {});
+});
```
noranda, line 15: TODO: add tests or remove.

packages/recipe/src/components/EzFormLabel/Implementations/EzFormLabelImplementation.tsx
```diff
@@ -0,0 +1,13 @@
+import React, {forwardRef} from 'react';
+import FormLabel from '@mui/material/FormLabel';
+import {EzFormLabelProps, Ref} from '../EzFormLabel.types';
+
+const EzFormLabelImplementation = forwardRef<Ref, EzFormLabelProps>((props, ref) => {
```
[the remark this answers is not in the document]
ez-alexfrazer: would I also change filename?

noranda (approved):
Looks great! Could you update the PR description and then this is good to go!

noranda (approved):

packages/recipe/src/components/EzFormLabel/EzFormLabel.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {forwardRef} from 'react';
+import {EzFormLabelMui} from './Implementations';
+import {EzFormLabelProps, Ref} from './EzFormLabel.types';
+
+const EzFormLabel = forwardRef<Ref, EzFormLabelProps>((props, ref) => {
```
danidewitt, line 5: do we want to be more explicit about some of the props we pass into this component? i'm thinking some of them should be, especially if we ever have other implementations (or if mui changes) @ez-alexfrazer @noranda

noranda: Not a bad idea to keep things in scope. @ez-alexfrazer I would look into which props you need to support for `EzRadio` and only include those. You can see how I did the same thing in [`EzChip`](https://github.com/ezcater/recipe/blob/main/packages/recipe/src/components/EzChip/EzChip.tsx#L6) to limit what props get passed in.

packages/recipe/src/components/index.ts
```diff
@@ -42,3 +42,4 @@ export {EzContent, EzHeader, EzItem, EzFooter, EzPreview} from './EzContent';
 export {EzPage, EzPageSection} from './EzPage';
 export {EzPageContent, EzContentGroup} from './EzPageContent';
 export {EzTimeline, EzTimelineEvent, EzTimelinePeriod} from './EzTimeline';
+export {default as EzFormLabel} from './EzFormLabel';
```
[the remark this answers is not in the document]
noranda: We don't want to expose this on the component grid because it's a util component, only used to support other components like `EzRadio`.

packages/recipe/src/components/EzFormLabel/EzFormLabel.tsx
noranda: @ez-alexfrazer Let's keep this simple and start with `children` as the only prop for this PR. If we need to add props (like `component`) later as part of `EzRadio`, we can do that then. It's a lot easier to add a prop than remove one.

EzFormLabel.jsx:
```js
...
const EzFormLabel = forwardRef<Ref, EzFormLabelProps>(({children}, ref) => (
  <EzFormLabelMui ref={ref}>{children}</EzFormLabelMui>
));
...
```

EzFormLabelMui.tsx:
```js
...
const EzFormLabelMui = forwardRef<Ref, EzFormLabelProps>(({children, ...props}, ref) => (
  <FormLabel {...props} ref={ref}>
    {children}
  </FormLabel>
));
...
```

EzFormLabel.types.ts:
```js
...
export interface EzFormLabelProps {
  children: ReactNode;
}
...
```

EzFormLabel.test.tsx:
```js
...
const Component: FC<Partial<ComponentProps<typeof EzFormLabel>>> = ({children}) => (
  <EzFormLabel>{children}</EzFormLabel>
);...
```

noranda: Changes pushed.

noranda (dismissed):

packages/recipe/src/components/EzFormLabel/EzFormLabel.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {forwardRef} from 'react';
+import {EzFormLabelMui} from './Implementations';
+import {EzFormLabelProps, Ref} from './EzFormLabel.types';
+
+const EzFormLabel = forwardRef<Ref, EzFormLabelProps>(({children}, ref) => (
```
danidewitt, line 5: ❓ do we minimally need focused, erorr, disabled, required here? what does this component do without those contexts?

another ❓ will we be able to use mui form label with implementations of child components that aren't mui?

noranda: @danidewitt

Q1: We may end up using those props when we implement EzRadio and/or EzCheckbox, but I think it's better to add those in when we've had time to decide exactly which are needed. Currently, the only required prop for this component is children.

Q2: I'd have to look at the source code to see exactly how it's used, but if we needed a new child implementation that didn't use MUI, we'd could always create new implementations of these control components if we needed to.

danidewitt (dismissed):

noranda (approved):

pgrippi (approved):

PR #845, opened 2022-06-23
make EzFlashMessage alert role

ez-alexfrazer:
## What did we change?

- makes EzFlashMessage into an alert role

## Why are we doing this?

Convenient for testing, more accessible too

```ts
// Code
function SaveProgressButton() {
  const {addFlashMessage} = useFlashMessages();
  const [saveProgress, {loading}] = useSaveProgressMutation({
    onCompleted() {
      addFlashMessage({
        type: 'success',
        headline: 'Your progress has been saved'
      });
    }
  });
  
  return (
    <EzButton
      use="primary"
      onClick={saveProgress}
      loading={loading}
    >
      {loading ? 'Saving...' : 'Save'}
    </EzButton>
  );
}
```

That would allow the following code to be used for testing
```ts
// Tests
test('shows flash message on save', async () => {
  render(<SaveProgressButton />);
  userEvent.click(await screen.findByRole('button', { name: /save/i }));
  expect(await screen.findByRole('alert', { name: /your progress has been saved/i })).toBeInTheDocument();
})
```

## Checklist


packages/recipe/src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -82,6 +83,24 @@ const icons = {
...
 
+/**
+ * Gets the appropriate role given the props passed in
+ * 1. if this is an info dialog, we probably don't need to aggressively alert them. For this, we use `role="status"`
+ * 2. if it is dismissable, we can use `role="alertdialog"`. "If the user is expected to close the alert, then the alertdialog role should be used instead."
+ * 3. in all other cases, assume a default of `role="alert"`
+ * @see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
+ * @returns an accessible role for the DOM element containing the alert.
+ */
+function getRole({use, onDismiss}: Pick<FlashMessageProps, 'use' | 'onDismiss'>) {
```
pgrippi, line 94: if you used `React.AriaRole` as a return type you shouldn't need the `as const` on the returns

```jsx
const getRole = ({use, onDismiss}: Pick<FlashMessageProps, 'use' | 'onDismiss'>): React.AriaRole => {
  if (use === 'info') {
    return 'status';
  }
  // ...
}
```

packages/recipe/src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -82,6 +83,24 @@ const icons = {
...
+  if (use === 'info') {
+    return 'status' as const;
+  }
+  if (typeof onDismiss !== 'undefined') {
+    return 'alertdialog' as const;
+  }
+  return 'alert' as const;
+}
```
noranda, line 102: This looks good (with the change @pgrippi suggested), however I'm not sure we should use `alertdialog` as a role here, where [initial focus goes to an element within the dialog](https://www.digitala11y.com/alertdialog-role/). Since we aren't using a modal/dialog here, I don't think this applies. It should probably just be `alert` if not `info` use. What do you think?

```suggestion
const getRole = ({use}: Pick<FlashMessageProps, 'use'>): AriaRole =>
  use === 'info' ? 'status' : 'alert';
```

You might just inline this now that it's this small.

ez-alexfrazer: Wouldn't we have to nest a ternary for that?

packages/recipe/src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -1,10 +1,11 @@
-import React, {useEffect} from 'react';
+import React, {type AriaRole, useEffect} from 'react';
```
noranda, line 1: Lint error: `'AriaRole' is defined but never used.`

noranda (dismissed):
Looks good! Thanks for your contribution!

packages/recipe/src/components/EzFlashMessage/EzFlashMessage.tsx
```diff
@@ -109,14 +111,24 @@ const EzFlashMessage: React.FC<FlashMessageProps> = ({
 
   if (!show) return null;
 
+  const role = use === 'info' ? 'status' : 'alert';
   return (
     <Style ruleset={theme}>
-      <div className={flashMessage({use})}>
+      <div
```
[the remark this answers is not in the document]
ez-alexfrazer: done

packages/recipe/src/components/EzFlashMessage/__tests__/EzFlashMessage.test.tsx
```diff
@@ -25,6 +25,34 @@ describe('EzFlashMessage', () => {
     expect(actual).toHaveNoViolations();
   });
 
+  describe('a11y', () => {
```
danidewitt, line 28: 🎉

danidewitt (approved):

noranda (approved):

PR #849, opened 2022-06-29
Add changelogs and titles for all changelogs

pgrippi:
## What did we change?

This adds a `CHANGELOG.md` file for every package and adds a title to the top of each changelog (just the package name)

## Why are we doing this?

I noticed that in the previous release and this upcoming one the version title inserted by the changeset tool is in the wrong location:

Screen Shot 2022-06-29 at 1 59 44 PM

I took a look at the changeset tool repo to see what they did, and each of their changelog files has an h1 title at the very top [(example)](https://raw.githubusercontent.com/changesets/changesets/main/packages/cli/CHANGELOG.md), so I've updated all of the changelog files here to match!

My running theory is that the changeset tool expects you to have a top-level title in the changelog doc, and is programmed to insert the next changelog under that title. Because it's missing in the recipe docs it thinks that the `##` header is the title, and puts anything new under that. By providing a static title at the top we can force it to put the next version info in the correct spot!

## Checklist


noranda (approved):
This looks good and makes sense! We recently removed the header from the recipe changelog, so that could explain this recent behavior. Thanks!

danidewitt (approved):

danidewitt:
thank you @pgrippi :tada

PR #850, opened 2022-06-30
workflow: …

noranda:
## What did we change?
This PR removes all deprecated snapshot images from Recipe and updates the contributor guide to reference chromatic.

## Why are we doing this?
As of `v15`, Recipe now uses [Chromatic](https://www.chromatic.com/) in place of jest snapshots and as such, all current snapshots in the codebase are deprecated and should be removed. This also updates the contributor guide to reference this change.

## Checklist


pgrippi (approved):

danidewitt (approved):
🥇

PR #851, opened 2022-06-30
Ez form control label

ez-alexfrazer:
## What did we change?

- Adds `EzFormControlLabel` component

## Why are we doing this?

Necessary component from mui

## Checklist


noranda (approved):
Looks good, just needs a rebase!

noranda (dismissed):

packages/recipe/src/components/EzFormControlLabel/EzFormControlLabel.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {forwardRef} from 'react';
+import {EzFormControlLabelMui} from './Implementations';
+import type {EzFormControlLabelProps, Ref} from './EzFormControlLabel.types';
+
+const EzFormControlLabel = forwardRef<Ref, EzFormControlLabelProps>((props, ref) => {
```
danidewitt, line 5: do we want to be more explicit about some of the props we pass into this component? i'm thinking some of them should be, especially if we ever have other implementations (or if mui changes) @ez-alexfrazer @noranda

.changeset/famous-beers-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': minor
+---
+
+[feat] Adds EzFormControlLabel component
```
danidewitt, line 5: i think this looks like `feat:` instead of `[feat]` for consistency in the changelog (unless that's changed - @noranda?)

packages/recipe/src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export {default as EzCarousel} from './EzCarousel';
 export {default as EzCheckbox} from './EzCheckbox';
 export {default as EzField} from './EzField';
 export {default as EzFlashMessage} from './EzFlashMessage';
+export {default as EzFormControlLabel} from './EzFormControlLabel';
```
danidewitt, line 12: are we going to expose this on the component grid? are we going to add documentation? does that also require a preview image? @noranda

.changeset/famous-beers-study.md
noranda: This has not changed and would be good for consistency.

packages/recipe/src/components/EzFormControlLabel/EzFormControlLabel.tsx
noranda: Not a bad idea to keep things in scope. @ez-alexfrazer I would look into which props you need to support for `EzRadio` and only include those. You can see how I did the same think in [`EzChip`](https://github.com/ezcater/recipe/blob/main/packages/recipe/src/components/EzChip/EzChip.tsx#L6) to limit what props get passed in.

packages/recipe/src/components/index.ts
noranda: We don't want to expose these ones on the component grid because they're util components, only used to support other components like `EzRadio`.

.changeset/famous-beers-study.md
noranda:
```suggestion
feat: adds EzFormControlLabel component
```

packages/recipe/src/components/EzFormControlLabel/EzFormControlLabel.tsx
noranda: @ez-alexfrazer Let's keep this simple and start with `control` and `label` as the only props for this PR. If we need to add additional props (like `disabled `, `labelPlacement`, or `value`) later as part of `EzRadio`, we can do that then. It's a lot easier to add a prop than remove one.

EzFormControlLabel.jsx:
```js
...
const EzFormControlLabel = forwardRef<Ref, EzFormControlLabelProps>(({control, label}, ref) => (
  <EzFormControlLabelMui control={control} label={label} ref={ref} />
));
...
```

EzFormControlLabelMui.tsx:
```js
...
const EzFormControlLabelMui = forwardRef<Ref, EzFormControlLabelProps>((props, ref) => (
  <FormControlLabel {...props} ref={ref} />
));
...
```

EzFormControlLabel.types.ts:
```js
...
export interface EzFormControlLabelProps {
  control: ReactElement<any, any>;
  label: ReactNode;
}
...
```

noranda: Changes pushed.

.changeset/famous-beers-study.md
noranda: Changes pushed.

noranda (dismissed):

danidewitt (approved):

noranda (approved):

PR #856, opened 2022-07-26
feat: Add EzBadge component

danidewitt:
## What did we change?
- Added a badge component

## Why are we doing this?
- Part of the EzNavigation epic and another domain's project

## Screenshot(s) / Gif(s):
Screen Shot 2022-07-26 at 2 05 30 PM
Screen Shot 2022-07-26 at 2 05 51 PM
Screen Shot 2022-07-26 at 2 06 00 PM
Screen Shot 2022-07-26 at 2 06 08 PM
Screen Shot 2022-07-26 at 2 06 18 PM
Screen Shot 2022-07-26 at 2 06 27 PM
Screen Shot 2022-07-26 at 2 06 32 PM

## Checklist

- [x] Provide test coverage for the changes made
- [x] Create a changeset (`yarn changeset`) with a summary of the changes

[Contributing guidelines](https://ezcater.github.io/recipe/guides/contributing/)

PR #857, opened 2022-08-16
EzBadge rework

noranda:
## What did we change?
- Reworked documentation to improve grammar, add additional examples, link to theme colors and contrast checker for accessibility, and wrap examples in `EzPage` to fix padding issues in the playroom
- Reworked component to:
  - remove `backgroundColor`, and `fontColor` props in favor of `color` which takes in a theme color
  - remove `variant` prop
  - add classes to MUI implementation for customization
- Fixed tests 
- Added regression tests
- Changed values from strings to numbers to fix issue with `max` and `showZero` not working correctly
- Fixed theme colors for `warning` in `ezTheme` and `EzChipMui`

## Why are we doing this?
Rework of `dd/ezbadge`: https://github.com/ezcater/recipe/pull/856


pgrippi (approved):

packages/recipe/src/components/EzBadge/EzBadge.md
```diff
@@ -25,32 +25,59 @@ Badges indicate notifications or events that are relevant and relatively close t
...
+      </EzBadge>
+    </EzPage>
   );
 };
 ```
 
 ### Badge Colors
 
-To use colors outside the variant options, provide a theme color to `backgroundColor` and `fontColor`. Ensure the contrast between the two is appropriate, consult with a designer if needed. Do not provide `variant` with `backgroundColor` and `fontColor`.
+EzBadge supports theme palette properties (`primary`, `secondary`, `error`, `warning`, `alert`, `info`, and `success`) as well as all supported [theme colors](http://localhost:8000/guides/theming/#colors) (ex. `common.red100`).
```
danidewitt, line 44: localhost

packages/recipe/src/components/EzBadge/EzBadge.tsx
```diff
@@ -3,53 +3,32 @@ import {EzBadgeMui} from './Implementations';
...
-        {children}
-      </EzBadgeMui>
-    );
-  }
+  ({alignX, alignY, children, color, hide, max, minimize, overlap, showZero, value}, ref) => (
+    <EzBadgeMui
+      ref={ref}
+      alignX={alignX}
+      alignY={alignY}
+      color={color}
```
danidewitt, line 11: 1. `color` in css is font color - which is why i kept the props for the interface `fontColor` and `backgroundColor` and created variants, instead of following what mui does (we may not have mui forever, so i wanted to future-proof it). 

2. is this going to make it harder for people to use custom font and background colors?

danidewitt: also, the props naming here is inconsistent with our other components - we use `variant` in other components. mui was inconsistent here also.

packages/recipe/src/components/EzBadge/EzBadge.md
noranda: Oops, good catch! Will fix.

packages/recipe/src/components/EzBadge/EzBadge.tsx
noranda: @danidewitt I understand where you're coming from, but I'd argue this is okay for a few reasons:
1. I agree we don't need to align with MUI props, but nor do we need to align with css props like `color`. In this case though, I think it is appropriate to use `color` here instead of `variant`, as only the color changes. I think `variant` better reflects cases with more than one difference in property. For example. `EzChip` uses `color` in the same way `EzBadge` does here when only the color changes, but it also uses `variant` for when multiple styles changes to outlines or to a specific status chip. In this case, `EzBadge` doesn't really have a different variant, just different colors.
2. We already use the `color` prop for this purpose in other MUI-implementation (see `EzChip` and `EzIcon`).

PR #861, opened 2022-08-30
Add EzRadio

noranda:
## What did we change?
- Adds `EzRadio` component to Recipe including MUI implementation, documentation, and regression tests.
- Adds `EzSuperFormControlLabel` component used internally in `EzFormControlLabel` when an `icon` is passed in to create a super radio button.
- Deprecates `EzSuperRadioButtons` and links to new `EzRadio` super radio buttons section.
- Deprecates `EzField` with `type="radio"` and links to new `EzRadio`.
- Updates `EzFormControlLabel` to add `disabled`, `helperText`, `icon`, `labelIcons`, and `value` props.
- Updates `EzFormLabel` to add `id` prop.
- Updates `EzRadioGroup` to add `ariaLabel`, `defaultValue`, `name`, `onChange`, `row`, and `value` props.
- Adds `WaterGlass` and `WineGlass` `EzIcon` mock icons.
- Adds `useThemeColor` theme hook to handle getting theme colors for use in components.
- Updates storybook and chromatic to hide Props stories.

## Why are we doing this?
This work implements the Radio Button redesign for both fullfillment and marketplace domains. Designs and implementations were reviewed and approved by appropriate designers.

Designs found [here](https://www.figma.com/file/mRXuf1mxiFCMOMWA2oKTdN/ezCater-DS?node-id=562%3A454).

## Screenshot(s) / Gif(s):
Screen Shot 2022-09-13 at 10 23 07 AM
Screen Shot 2022-09-13 at 10 23 38 AM
Screen Shot 2022-09-13 at 10 23 51 AM
Screen Shot 2022-09-13 at 10 24 01 AM
Screen Shot 2022-09-13 at 10 24 12 AM

## Checklist


packages/recipe/src/components/EzFormControlLabel/Implementations/EzFormControlLabelMui.tsx
```diff
@@ -1,10 +1,77 @@
...
+import EzSuperFormControlLabel from '../../EzSuperFormControlLabel';
+import {useThemeColor} from '../../../themes/hooks/useThemeColor';
+import {EzIconTypes} from '../../EzIcon/EzIcon.types';
 
-const EzFormControlLabelMui = forwardRef<Ref, EzFormControlLabelProps>((props, ref) => (
-  <FormControlLabel {...props} ref={ref} />
-));
+const EzFormControlLabelMui = forwardRef<Ref, EzFormControlLabelProps>(
+  ({control, disabled, helperText, icon, label, labelIcons, ...props}, ref) => {
+    if (icon)
```
pgrippi, line 10: this `if` should be moved below the `useTheme` / `useThemeColor` hooks below (I'm surprised eslint didn't catch this)

hooks need to fire in the same order every render, which means you cannot early return before calling them

packages/recipe/src/components/EzFormControlLabel/Implementations/EzFormControlLabelMui.tsx
```diff
@@ -1,10 +1,77 @@
...
+        classes={{root: 'EzFormControlLabel', label: 'EzFormControlLabel-label'}}
+        control={control}
+        disabled={disabled}
+        label={
+          <Stack lineHeight="16px">
+            <Stack direction="row">
+              {label}
+
+              {labelIcons &&
+                labelIcons.map((labelIcon: EzIconTypes, index: number) => (
```
pgrippi, line 37: you can use the safety navigator operator here:

```jsx
{labelIcons?.map(() => ...)}
```

packages/recipe/src/components/EzFormLabel/EzFormLabel.types.ts
```diff
@@ -4,4 +4,5 @@ export type Ref = HTMLLabelElement;
 
 export interface EzFormLabelProps {
   children: ReactNode;
+  id: string;
```
pgrippi, line 7: is this a breaking change, requiring an ID on form labels?

packages/recipe/src/components/EzSuperFormControlLabel/Implementations/EzSuperFormControlLabelMui.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const theme = useTheme();
+    const themeColor = control.props.color?.startsWith('common.')
+      ? theme.palette.common[control.props.color.split('.')[1]]
+      : theme.palette[control.props.color].main;
+    const themeColorDark = darken(themeColor, 0.5);
+    const disabledColor = theme.palette.common.neutral140;
+
+    return (
+      <Stack
+        ref={ref}
```
pgrippi, line 22: 🎱 should this ref be passed to the `FormControlLabel` instead of the stack? I would expect the ref on a control label to be the label element

packages/recipe/src/components/EzSuperRadioButtons/EzSuperRadioButtons.md
```diff
@@ -6,6 +6,16 @@ path: '/components/ez-super-radio-buttons'
 tags: ['radio']
 ---
 
+<EzAlert
+  headline="Deprecation Warning"
+  tagline="EzSuperRadioButtons is deprecated as of v15 and will be removed in v16. Super Radio Buttons can be used with EzRadio."
```
pgrippi, line 11: 🎱 would it be worth adding a deprecation warning in the console when this component mounts? We had done something similar with EzIcon logging a warning when using the old icon sizes

packages/recipe/src/components/EzRadio/Implementations/EzRadioMui.tsx
```diff
@@ -0,0 +1,142 @@
+import React, {forwardRef} from 'react';
+import {Radio, Stack, useTheme, Zoom} from '@mui/material';
+import {EzRadioProps, Ref} from '../EzRadio.types';
+import {useThemeColor} from '../../../themes/hooks/useThemeColor';
+import {PaletteOptions} from '../../../themes/themes.types';
+
+const VARIANT_SIZES = {
```
pgrippi, line 7: it's too bad MUI doesn't easily allow us to override variants on components that don't take a variant prop, so much styling has to be manually set in code...

pgrippi (commented):
I have an unrelated question, I've noticed that all of our new components are setting `defaultProps`. I kinda forgot that existed, and when [I looked at the docs](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) it's clearly to support class components

From what I've seen with modern react using anonymous functions, it's preferred to write components like:

```jsx
const Foo = ({myProp, myDefaultedProp = false}) => ( ... );
```

What are your y'alls thoughts on migrating away from `defaultProps`?

noranda:
> What are your y'alls thoughts on migrating away from `defaultProps`?

@pgrippi We've been using `defaultProps` for consistency and have it baked into our `create-component` boilerplate components. Based on my initial research, there is not a significant performance difference, however it does look like `defaultProps` for functional components [will be deprecated in the future](https://twitter.com/dan_abramov/status/1133878326358171650) ([RFC](https://github.com/reactjs/rfcs/pull/107)), so we should discuss this change, although out of scope for this PR as it should be applied across Recipe.

packages/recipe/src/components/EzFormControlLabel/Implementations/EzFormControlLabelMui.tsx
noranda: Nice catch!

packages/recipe/src/components/EzFormControlLabel/Implementations/EzFormControlLabelMui.tsx
noranda: ✅

packages/recipe/src/components/EzSuperFormControlLabel/Implementations/EzSuperFormControlLabelMui.tsx
noranda: This makes sense! In fact, I don't think we even need to wrap this in a `Stack` component. I will rework.

packages/recipe/src/components/EzFormLabel/EzFormLabel.types.ts
noranda: Technically not, because we haven't released these new supporting components yet. :)

packages/recipe/src/components/EzSuperRadioButtons/EzSuperRadioButtons.md
noranda: I think this is a great idea. I will add.

packages/recipe/src/components/EzRadio/Implementations/EzRadioMui.tsx
noranda: Yeah, I wasn't thrilled with this approach, but it does keep all the variant styles together instead of embedding them in the component, making it easier to add/remove/edit variants in the future.

pgrippi (approved):

pgrippi (approved):

packages/recipe/src/components/EzRadio/EzRadio.md
```diff
@@ -0,0 +1,489 @@
...
+    name: 'basic-radio-button',
+    onChange: handleChange,
+    value: item,
+  });
+
+  return (
+    <EzLayout>
+      <EzRadio {...radioProps('a')} />
+
+      <EzLayout style={{backgroundColor: '#034a34'}}>
```
danidewitt, line 44: i'm sure there's a reason for these inline style (and the ones below it), it's just not super obvious to me. i'd generally lean towards not doing this so people don't copy things we've done. thoughts?

packages/recipe/src/components/EzRadio/EzRadio.md
```diff
@@ -0,0 +1,489 @@
...
+        </EzFormControl>
+      </EzLayout>
+    </EzLayout>
+  );
+};
+```
+
+---
+
+## Custom Styles
```
danidewitt, line 406: i'm on the fence about exposing this - is it really going to be used? and is that optimal?

danidewitt (dismissed):
i only have 2 questions - all related to documentation. great job on this!

packages/recipe/src/components/EzRadio/EzRadio.md
noranda: @danidewitt No good reason, other than I needed a darker background color. I'll change these to `<div>`s just so there isn't any confusion.

packages/recipe/src/components/EzRadio/EzRadio.md
noranda: @danidewitt I talked about this with @JHojberg when I worked on `EzChip` and there are cases where our components will need to be customized, which is why we expose these classes. We do state that users should use supported styles and only use these when absolutely necessary. This is how MUI handles custom styling as well.

pgrippi (approved):

packages/recipe/src/components/EzRadio/EzRadio.md
noranda: Done!

packages/recipe/src/components/EzRadio/EzRadio.md
danidewitt: yeah, that makes sense. the only difference with comparing to mui is that mui isn't an opinionated design system in the same way recipe is.

noranda: True! I think this makes sense for us though after discussions with design. We may want to revisit this later.

PR #864, opened 2022-09-28
Fix a memory leak and support using React 18's useId hook (if present)

pgrippi:
## What did we change?

This addresses a memory leak we're seeing in our production apps that use Recipe. The `useUniqueId` method is holding onto the `setState` method returned from `useState` that is never cleared during an SSR render cycle, leading to an object that grows infinitely

This also updates the `useUniqueId` hook so that it uses the React 18 `useId` hook if present, falling back to our custom implementation when using React 17 or lower.

As part of this I had to remove the `mergeIds` method. I don't know what this might break, worse case it'll always take the ID from the last object passed to the function (mimicking what `Object.assign` would do)

## Why are we doing this?

* Memory leaks are bad
* In order for store to upgrade to react 18 we need to fix hydration errors, which are happening with any component using the custom `useUniqueId` hook (the IDs will always be mismatched between the SSR and hydration cycle)

## Screenshot(s) / Gif(s):

The memory leak we're seeing:

Screen Shot 2022-09-28 at 10 48 35 AM

## Checklist


onocy (approved):
I'm curious how you were able to figure out that this was the issue? I would love any insight on that exploration

pgrippi:
> I'm curious how you were able to figure out that this was the issue? I would love any insight on that exploration

Sadly I'm not sure this is the issue (I'm not sure there is just one memory leak)! BUT it is definitely a memory leak - we are holding onto these functions and never deleting them (`useEffect` will delete them but it never runs on the server side)

I found this by accident really, I came in here originally to just swap this out with React 18's `useid` and stumbled upon us storing these methods in a global variable...

noranda (approved):
Looks good! All test pass, no changes in Chromatic. I did look into the [PR](https://github.com/ezcater/recipe/pull/403) where the `mergeIds` method was added, and it was related to selects, which do not appear to have changed as a result of this PR.

PR #866, opened 2022-09-30
Fix warning around `useId` not being exported from react

pgrippi:
## What did we change?

This updates the guard around `React.useId` being undefined (running react 17)

## Why are we doing this?

Webpack is complaining about `useId` not being defined:

```
./node_modules/@ezcater/recipe/dist/esm/utils/hooks/useUniqueId.js
Attempted import error: 'useId' is not exported from 'react' (imported as 'React').
```

I originally thought that doing `import * as React from 'react'` was the way around this but it turns out webpack is smart enough to still complain about it!

MUI has implemented their own `useId` hook that attempts to use React 18's `useId` hook and falls back to a custom solution when using React 17 - exactly what we want to try here - so I [used their implementation](https://github.com/mui/material-ui/blob/d1ae04f3adffce5c499f7b154237f7fc05452cc0/packages/mui-utils/src/useId.ts#L21) where they solved the webpack problem!

I did not include a changelog for this since we haven't released yet.

## Checklist


packages/recipe/src/utils/hooks/useUniqueId.ts
```diff
@@ -5,12 +5,11 @@ const getUniqueId = (() => {
   return () => `control__${index++}`;
 })();
 
-const useCustomUniqueId = (): string => {
-  return React.useMemo(getUniqueId, []);
-};
+// eslint-disable-next-line no-useless-concat -- Workaround for https://github.com/webpack/webpack/issues/14814
+const maybeReactUseId: undefined | (() => string) = (React as any)['useId' + ''];
```
pgrippi, line 9: This is a fancy way to trick webpack! `React['useId']` is the same as `React.useId`, both of which webpack will see and complain about. But as soon as you throw in a string concatenation webpack is no longer able to statically analyze the import and complain about it not existing! 😂

pgrippi:
I created a beta tag ([0.0.0-canary-20220930135217](https://github.com/ezcater/recipe/releases/tag/%40ezcater/recipe%400.0.0-canary-20220930135217)) and confirmed on store-next that the build error is no longer present after this change, we're all set to go 🎉

noranda (approved):

evan-007 (approved):

onocy (approved):

PR #869, opened 2022-10-03
Fix the `useId` React 18 feature detection in a way that works with esbuild-loader

pgrippi:
## What did we change?

This further obfuscates the `useId` import so that it works with esbuild, webpack, and swc for apps using React 17

## Why are we doing this?

Follow up to https://github.com/ezcater/recipe/pull/866 that worked for webpack/swc but not esbuild

## Checklist


danidewitt (approved):

onocy (approved):

pgrippi:
Verified this still works in store-next without errors, :shipit: !

PR #876, opened 2022-10-31
v16 Deprecations

noranda:
## What did we change?
This PR removes the following deprecated Recipe components:
- `EzStatus` (use `EzChip`)
- `EzPageContent` (use `EzPage`)
- `EzContentGroup` (use `EzLayout`)
- `EzSuperRadioButtons` (use `EzRadio`)
- `EzLabelledItem` (use `EzLabel`)

Note: `EzField` with `type="radio"` was scheduled to be removed in v16, but was pushed back to v17 to give users more time to update and to coincide with `EzField` with `type="checkbox"` which is deprecated as of v16 and also scheduled to be removed in v17.

Updated v16 Migration Guide.

## Why are we doing this?
Cleaning up deprecated components. There were a few deprecated properties and types of some of our components that I chose to leave as-is, since we plan to refactor those components in the near future.

## Checklist


danidewitt (approved):
1️⃣ 6️⃣ 🎉

pgrippi (approved):

PR #877, opened 2022-11-04
V16

noranda:
This PR is to merge v16 into main in preparation for the Recipe 16.0.0 release. Branches are in sync.


packages/recipe/src/components/EzFormGroup/EzFormGroup.types.ts
```diff
@@ -0,0 +1,9 @@
+import {ReactNode} from 'react';
+
+export type Ref = HTMLDivElement;
+
+export interface EzFormGroupProps {
+  ariaLabel?: string;
```
danidewitt, line 6: ❓ why is aria label optional? should it be required?

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
```diff
@@ -0,0 +1,50 @@
...
+) & {
+  /** @deprecated */
+  className?: string;
+  /** @deprecated */
+  label?: string;
+  onChange?: (event: ChangeEvent<HTMLInputElement>, checked: boolean) => void;
+};
+
+export type EzCheckboxMuiProps = EzCheckboxCommonProps & {
+  ariaLabel?: string;
```
danidewitt, line 39: same question about optional aria label

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
```diff
@@ -0,0 +1,50 @@
...
+  color?: EzThemeColors;
+  defaultChecked?: boolean;
+  onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
```
danidewitt, line 42: does it always need an onChange?

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
```diff
@@ -0,0 +1,50 @@
...
+  size?: string;
+  value?: any;
```
danidewitt, line 44: does it always need a value?

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
```diff
@@ -0,0 +1,50 @@
...
+export type EzCheckboxLegacyAcknowledgementProps = EzCheckboxCommonProps & {
+  /** @deprecated */
+  acknowledgement: true;
+  /** @deprecated */
+  terms: string;
+};
+
+export type EzCheckboxLegacyClassicProps = EzCheckboxCommonProps & {
+  acknowledgement?: boolean;
+  terms?: undefined;
```
danidewitt, line 24: why undefined here?

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
```diff
@@ -0,0 +1,50 @@
+import {ChangeEvent, FocusEventHandler} from 'react';
+import {type EzThemeColors} from '../../themes';
+
+export type Ref = HTMLInputElement;
+
+export type EzCheckboxCommonProps = {
+  id?: string;
```
danidewitt, line 7: does it always need an id?

danidewitt:
👍 looks great, as it should since this was all code reviewed before. i had a couple props questions after re-reviewing, but nothing major.

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
noranda: If the checkbox has a label, then `aria-label` isn't required.

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
noranda: It only needs an `onChange` if it's controlled.

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
noranda: `value` defaults to "on" ([source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value))

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
noranda: In this case it's similar to `never` and was needed to type narrow it from the union type. Typescript is weird!

packages/recipe/src/components/EzCheckbox/EzCheckbox.types.ts
noranda: As long as the checkbox is wrapped inside the label (as ours are), the `id` is optional.

```js
// id optional
<label>
  <input type="checkbox">
  text
</label>

// id required
<label htmlFor="cb-1">text</label>
<input type="checkbox" id="cb-1">
```

pgrippi (approved):

danidewitt (approved):

PR #880, opened 2022-11-15
Update ezAlert types to explicitly allow ReactNode

onocy:
## What did we change?
- Updated types of EzAlert

## Why are we doing this?

We would like to be able to mock the following feature (mobile customer message alert with link) by using recipe's `EzAlert`. 
Screen Shot 2022-11-11 at 9 54 10 AM

## Screenshot(s) / Gif(s):

`EzAlert` accepts markdown as props implicitly and the code renders (as you can [see in this playground)](https://ezcater.github.io/recipe/playroom/?code=N4Igxg9gJgpiBcIA8BRAXgQQDYwE4BcACACxgEMosBLAOxgF4AdEAZTyogFcBnQgdzK4atAOa8IuQnlwTuzQvjIjqdesCRkSuGADMmIAMTyAfAE0uhOjCgKIhHVQAehbhAC2MfMVEA6JAHoyYwBfQh4GZmkJeX9jEAAaED4qKC9uBABtAGYAJgAGAF1goA) but we want to update the types here so that the linter doesn't complain. 

Screen Shot 2022-11-15 at 10 40 38 AM

## Checklist

- [X] Create a changeset (`yarn changeset`) with a summary of the changes


onocy:
I may have missed some step as far as contributing given the size of this change and that it has been a while. Any suggestions are welcome 😄

amberlcook (dismissed):
🎊

onocy:
> This branch is 118 commits behind `main` including our change to yarn workspaces. Do you need this fix on an earlier version? Or can this be rebased on `main`?

Yeah I didn't pull main before creating this branch I'll do that now

pgrippi (dismissed):

packages/recipe/src/components/EzAlert/EzAlert.tsx
```diff
@@ -15,7 +15,7 @@ const icons = {
 type Props = {
   arrow?: 'top' | 'bottom';
   headline: string;
-  tagline?: string | Element;
+  tagline?: string | Element | React.ReactNode;
```
noranda, line 18: Lets get rid of `Element` and replace it with `ReactNode` here (and import it at the top).
```suggestion
  tagline?: string | ReactNode;
```

noranda (changes requested):
This is an older component that does have type checking in it's test. Let's add it!

Add this to `/EzAlert/__tests__/EzAlert.test.tsx`:
```js
  it('should pass type checking', () => {
    [
      {
        default: <EzAlert headline="Alert" />,
        tagline: <EzAlert headline="Alert" tagline="Tagline" />,
        success: <EzAlert use="success" headline="Alert" />,
        error: <EzAlert use="error" headline="Alert" />,
        warning: <EzAlert use="warning" headline="Alert" />,
        info: <EzAlert use="info" headline="Alert" />,
        tip: <EzAlert use="tip" headline="Alert" />,
        marketing: <EzAlert use="marketing" headline="Alert" />,
        topArrow: <EzAlert arrow="top" headline="Alert" />,
        bottomArrow: <EzAlert arrow="bottom" headline="Alert" />,
        reactNodeTagline: <EzAlert headline="Alert" tagline={<div>Alert</div>} />,
      },
    ].forEach(() => {});

    expect.assertions(0);
  });
```

packages/recipe/src/components/EzAlert/EzAlert.tsx
onocy: Part of the reason why I did React.ReactNode here was that the import without direct usage was considered to be a linting error. I'm not sure why the linter wasn't picking up the type definition as "usage", but this was my way of skirting around having to define that import with the lint ignore line.

noranda: Strange! Seems fine on my end, but I'm not worried about it. 👍

pgrippi (approved):
:shipit:

noranda:
Thanks @onocy for your contribution!! 🥇

PR #882, opened 2022-11-28
Update EzRadio and EzRadioGroup props to match MUI

pgrippi:
## What did we change?

This updates the typescript signature for the `onChange` callback for the `EzRadio` and `EzRadioGroup` to match MUI

## Why are we doing this?

I ran into this trying to pass an onChange method to `EzRadioGroup` when upgrading store-next. In order to get the value and checked status I need to do `event.target.value`/`event.target.checked`, but the typescript types in recipe doesn't expose an event so I get a typescript error. Functionally it works so it's an easy fix!

Ref that I used:

- [Radio props](https://mui.com/material-ui/api/radio/#props)
- [RadioGroup props](https://mui.com/material-ui/api/radio-group/#props)

For good measure I checked the `EzCheckbox` props and they match what MUI uses so I think we're good there!

## Checklist


packages/recipe/src/components/EzRadioGroup/EzRadioGroup.types.ts
```diff
@@ -7,7 +7,7 @@ export interface EzRadioGroupProps {
   children: ReactNode;
   defaultValue?: any;
   name?: string;
-  onChange?: () => any;
+  onChange?: (event: ChangeEvent<HTMLInputElement>, value: string) => void;
```
pgrippi, line 10: I don't know why the `onChange` for the radio group has a second argument passed to it, neither the checkbox nor radio components get it.

I double checked the source code and sure enough, they pass in `event.target.value` as the second argument: https://github.com/mui/material-ui/blob/1ac8b2107baa0c7050133628adfd27e1e2f721db/packages/mui-material/src/RadioGroup/RadioGroup.js#L58

But why is a mystery! Seems like the consumer could just do `event.target.value`...

danidewitt (approved):
thanks @pgrippi 🥇

noranda (approved):
Looks good, thanks!

PR #885, opened 2022-11-30
Update legacy checkbox and radio buttons to log the deprecated message once

pgrippi:
## What did we change?

This updates the legacy radio button and checkbox button components so that they log the deprecation message the first time they mount rather than every time they render

## Why are we doing this?

I noticed this when I was working on upgrading to the new components in store-next, my console was getting slammed with deprecation warnings even though there were only a few of them on the page. I think it's good to log it once to get the stack trace, which helps figure out where the component is being rendered, but it doesn't need to happen more than once!

## Checklist


danidewitt (approved):

royce-ezcater (approved):
Seems reasonable to me!

noranda (approved):

PR #887, opened 2022-12-12
fix: remove redundant aria-labels from EzRadio and EzCheckbox

noranda:
## What did we change?
Removed redundant aria-labels from both `EzRadio` and `EzCheckbox` as follows:
- an `aria-label` was duplicated on both the input and parent elements - removed from the parent element
- an `aria-label` was unnecessarily applied to inputs that were wrapped in label elements - removed

## Why are we doing this?
[Aria-labels](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) are necessary for accessibility, but were not being used appropriately here. The `aria-label` should be applied directly to the input and only when necessary (i.e. not wrapped in a `<label>` element).

_Note: if a user wants to include an `aria-label` in a label wrapped input, they still can, but it is no longer done by default._

### Before - EzRadio:
Screen Shot 2022-12-12 at 1 09 43 PM
Screen Shot 2022-12-12 at 1 10 46 PM

### Before - EzCheckbox:
Screen Shot 2022-12-12 at 1 11 33 PM
Screen Shot 2022-12-12 at 1 12 17 PM

### After - EzRadio:
Screen Shot 2022-12-12 at 1 13 17 PM
Screen Shot 2022-12-12 at 1 13 50 PM
Screen Shot 2022-12-12 at 1 14 15 PM

### After - EzCheckbox:
Screen Shot 2022-12-12 at 1 14 36 PM
Screen Shot 2022-12-12 at 1 15 07 PM
Screen Shot 2022-12-12 at 1 15 33 PM

## Checklist


packages/recipe/src/components/EzCheckbox/Implementations/EzCheckboxMui.tsx
```diff
@@ -88,7 +88,6 @@ const EzCheckboxMui = forwardRef<Ref, EzCheckboxMuiProps>(
     return (
       <Checkbox
         inputRef={ref}
-        aria-label={ariaLabel}
```
pgrippi, line 91: 🎱 isn't this still needed?

noranda: You got it!

packages/recipe/src/components/EzFormControlLabel/Implementations/EzFormControlLabelMui.tsx
```diff
@@ -14,7 +14,7 @@ const EzFormControlLabelMui = forwardRef<Ref, EzFormControlLabelProps>(
       return (
         <EzSuperFormControlLabel
           ref={ref}
-          control={control}
+          control={cloneElement(control, {ariaLabel: control.props.ariaLabel || null})}
```
danidewitt, line 17: is there a case where it is ever null? should it ever be null?

danidewitt: disregard i just re-read the pr description!

danidewitt (approved):

PR #888, opened 2022-12-15
style: add optional classname suffix prop to EzField and associated EzPopper

noranda:
## What did we change?
- Adds `"EzField"` classname to `<EzField>`.
- Adds `"EzPopper"` classname to `<EzPopper>`.
- Adds the optional `classNameSuffix` prop to `EzField`, which when provided will do the following:
  - Adds `"EzField-[classNameSuffix]"` to `<EzField>`.
  - Adds `"EzPopper-[classNameSuffix]"` to `<EzPopper>` for `<EzField>` inputs with types `select`, `autosuggest`, and `date`.

## Why are we doing this?
`EzListBox` [adds a min-width](https://github.com/ezcater/recipe/blob/main/packages/recipe/src/components/EzField/EzListBox.tsx#L8) of `230px`, which is causing some issues when using `EzField` inputs on mobile when they are smaller than `230px`. Additionally, `EzPopper`, which uses the now unsupported `@popperjs/core` library, has a bug which adds an offset to the popper under certain circumstances, causing layout issues. As we are planning to refactor these inputs in the near future to use MUI implementations, these issues should be fixed, but in the meantime users need to be able to access these components to fix the styles.

## Checklist


pgrippi (approved):
:shipit:

jmargolisvt (approved):
🚀

PR #895, opened 2023-02-08
allow_for_ezcard_heading_with_only_actions

paul-ez:
## What did we change?
Renders EzTable with header when `actions` are passed OR `title`

## Why are we doing this?
Currently table will only render a header if `title` is passed. Our current design for updating ezmange Rankings ui requires this.

## Checklist


danidewitt:
i know this is still a draft - so i'm commenting before you go too far down the path of more work. our [best practices for table](https://ezcater.github.io/recipe/components/ez-table#best-practices) include a title "include a heading to help identify the contents of the table at-a-glance" - can we confirm with design to make sure a table is still going to be accessible without a title?

PR #897, opened 2023-02-13
Add callback function to `EzTooltip` 

onocy:
## Checklist


onocy:
Do I need an account with chromatic to be able to approve this? The diff looks good to me 
Screen Shot 2023-02-13 at 4 52 06 PM

packages/recipe/src/components/EzTooltip/__tests__/EzTooltip.test.tsx
```diff
@@ -49,6 +49,18 @@ describe('EzTooltip', () => {
...
+      const {container} = render(
+        <EzTooltip message="tooltip visible" position="horizontal" onShowTooltip={onShowTooltip}>
+          <input />
+        </EzTooltip>
+      );
+
+      fireEvent.mouseEnter(container.querySelector('input'));
+
+      expect(onShowTooltip).toHaveBeenCalled();
+    });
```
pgrippi, line 63: 🎱 might be overkill, but maybe add another test for `focus`? That would exercise both boolean cases for `showTooltip`

packages/recipe/src/components/EzTooltip/EzTooltip.tsx
```diff
@@ -66,6 +67,12 @@ const EzTooltip: React.FC<Props> = ({children, message, position, ...rest}) => {
   const showTooltip = isHovered || isFocused;
   const child = React.Children.only(children);
 
+  useEffect(() => {
+    if (showTooltip && onShowTooltip) {
+      onShowTooltip();
+    }
+  }, [showTooltip, onShowTooltip]);
```
pgrippi, line 74: 🎱 this may be one of those over optimizations, but I wonder if we should stick the callback in a ref and use that rather than have it as a dependency in this useEffect hook

I imagine the use case of this will look like:

```tsx
<EzTooltip message="..." onShowTooltip={() => trackingMethodCall()} />
```

It is unlikely that `onShowTooltip` will be a memoized function (and it doesn't really need to be), but if something causes that component using EzTooltip to re-render while the tooltip is open, `onShowTooltip` will be a different function reference and it will re-trigger the callback. It probably isn't a huge deal, but it could be avoided by doing something like this:

```tsx
// this is the `useLatest` hook from react-use
const onShowTooltipRef = useRef(onShowTooltip);
onShowTooltipRef.current = onShowTooltip;

useEffect(() => {
  if (showTooltip) onShowTooltipRef.current?.();
}, [showTooltip, onShowTooltipRef]);
```

This will cause the `useEffect` hook to only ever fire when `showTooltip` changes, with the understanding that passing in a different function won't matter until the tooltip re-opens (which, my hunch tells me is the desired behavior)

pgrippi (dismissed):

noranda:
> Do I need an account with chromatic to be able to approve this? The diff looks good to me

@onocy This is a known threshold issue in Chromatic at the moment and can be safely approved. Your latest build reverted it, so all good!

noranda (changes requested):
This looks good with the changes that @pgrippi suggested. Thanks for your contribution!

packages/recipe/src/components/EzTooltip/EzTooltip.tsx
onocy: updated in [77151ad](https://github.com/ezcater/recipe/pull/897/commits/77151ad3f8bf7544a684f00365dfabc51ca81d68)

packages/recipe/src/components/EzTooltip/__tests__/EzTooltip.test.tsx
onocy: updated in [77151ad](https://github.com/ezcater/recipe/pull/897/commits/77151ad3f8bf7544a684f00365dfabc51ca81d68)

.changeset/twelve-trees-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ezcater/recipe': minor
+---
+
+Add callback function to EzTooltip for tracking
```
noranda, line 5: Oops, can you also change this to the following to keep with our changelog format?
```suggestion
feat: add callback function to EzTooltip for tracking
```

onocy: Yes sorry about that. I used `yarn changeset` to generate this file I believe, it's too bad you can't pick amongst a list of options to prepend to the description (unless it's there and I missed it).

pgrippi (approved):

.changeset/twelve-trees-bake.md
noranda: It is not, but that's a good idea! I might look into that. Otherwise, the options are listed on our [doc site](https://ezcater.github.io/recipe/guides/versioning-and-changelog). Thanks again!

noranda (approved):
Nice work!

PR #898, opened 2023-02-14
Tweak the changesets and yarn workspace settings so we can remove our custom versioning script

pgrippi:
## What did we change?

This PR updates the `package.json` files for our internal packages to use `*` as the package version, which allows `yarn` to correctly link to the internal version of the package after its version has been updated.

This then allows us to remove the `sync-package-versions` script, which was made to update the recipe and other package versions. The cascading effect is that internal packages that rely on recipe no longer need to be changed on a release, preventing the changesets tool from seeing a change and thinking it needs to release a new version.

## Why are we doing this?

On `main`, the changeset tool thinks that we need to release a `@ezcater/theme-marketplace` major version, even though nothing has changed in it:

```
❯ yarn changeset status
yarn run v1.22.19
$ /Users/petergrippi/Projects/recipe/node_modules/.bin/changeset status
🦋  info NO packages to be bumped at patch
🦋  ---
🦋  info Packages to be bumped at minor:
🦋  info 
🦋  - @ezcater/recipe
🦋  - @ezcater/recipe-doc-site
🦋  ---
🦋  info Packages to be bumped at major:
🦋  info 
🦋  - @recipe-ui/theme-marketplace
```

After my changes, it correctly only sees two things that need to be bumped - the two packages that actually had their versions bumped via changesets:

```
❯ yarn changeset status
yarn run v1.22.19
$ /Users/petergrippi/Projects/recipe/node_modules/.bin/changeset status
🦋  info NO packages to be bumped at patch
🦋  ---
🦋  info Packages to be bumped at minor:
🦋  info 
🦋  - @ezcater/recipe
🦋  - @ezcater/recipe-doc-site
🦋  ---
🦋  info NO packages to be bumped at major
✨  Done in 0.47s.
```

## Caution

I will warn ya'll that this seems to work locally, but if it does end up breaking `yarn workspaces` it's likely we won't see that until after we merge this and cut a new release.

If that happens I'll hop on and keep playing around with it!


packages/recipe/src/utils/domProps.ts
```diff
@@ -10,7 +10,7 @@ import {mergeProps} from './mergeProps';
  */
 export const domProps = (
   props: unknown,
-  ...styledExpressions: ReturnType<typeof stitches['createTheme']>[]
+  ...styledExpressions: ReturnType<(typeof stitches)['createTheme']>[]
```
pgrippi, line 13: it looks like when I upgraded `@changesets/cli`, `prettier` came long for the ride. Looking at the `yarn.lock` changes it also looked like recipe had two versions of prettier installed: a 1.0 and a 2.0 version

I have bumped the version properly in package.json and fixed this one issue it found

noranda (approved):

noranda:
I'm going to merge this with one approval since @danidewitt is out and this doesn't change anything in Recipe core. Plus I really want to test it out!

PR #907, opened 2023-03-01
fix: remove default ariaLabel prop from EzRadio to fix overriding wrapped label

noranda:
## What did we change?
- Removed default `ariaLabel = 'radio button'` from `EzRadio`.
- Added console warning if a radio button is ever used without a wrapped label and without an `ariaLabel`.
- Added accessibility tests.

## Why are we doing this?
The default `ariaLabel` on `EzRadio` was overriding wrapped labels, which we don't want. Removing the default can lead to the radio button being inaccessible, so a console warning and accessibility tests were added to prevent this.

## Screenshot(s) / Gif(s):
Before:
Screenshot 2023-03-01 at 9 52 21 AM

After:
Screenshot 2023-03-01 at 9 50 43 AM

Console Warning:
Screenshot 2023-03-01 at 9 51 38 AM

## Checklist


packages/recipe/src/components/EzRadio/EzRadio.types.ts
```diff
@@ -1,7 +1,7 @@
 import {ChangeEvent} from 'react';
 import {type EzThemeColors} from '../../themes';
 
-export type Ref = HTMLDivElement;
+export type Ref = HTMLInputElement;
 
 export interface EzRadioProps {
   ariaLabel?: string;
```
danidewitt, line 7: so we can't make aria label here required... because the label would conflict with it? or something else?

noranda: To be accessible, a radio button either needs to have an `ariaLabel` or wrapped in a `label` element. So `ariaLabel` is not required here.

danidewitt (approved):

noranda:
Rebased this off `main` so we can get this into `v16`. Needs rereview @pgrippi and @danidewitt. Thanks!

pgrippi (approved):

danidewitt (approved):

PR #921, opened 2023-03-15
feat: add transparent background color prop to EzCard

danidewitt:
## What did we change?
- Add `transparent` flag to `EzCard`

## Why are we doing this?
- Designer request

## Screenshot(s) / Gif(s):
Screen Shot 2023-03-15 at 1 17 50 PM

## Checklist

- [x] Provide test coverage for the changes made
- [x] Create a changeset (`yarn changeset`) with a summary of the changes

[Contributing guidelines](https://ezcater.github.io/recipe/guides/contributing/)

PR #922, opened 2023-03-15
feat: add transparent background color prop to EzTable

danidewitt:
## What did we change?
- Add `transparent` flag to `EzTable`
- ⚠️ Should be merged into https://github.com/ezcater/recipe/pull/921 before merging https://github.com/ezcater/recipe/pull/921

## Why are we doing this?
- Designer request

## Screenshot(s) / Gif(s):
Screen Shot 2023-03-15 at 1 21 40 PM

## Checklist


paul-ez (approved):

packages/recipe/src/components/EzTable/__tests__/EzTable.test.tsx
```diff
@@ -389,6 +389,13 @@ describe('EzTable', () => {
     });
   });
 
+  it('should display a table with a transparent background', async () => {
+    const {getAllByText} = render(<EzTable columns={columns} items={items} transparent />);
+    expect(getAllByText('Store name')[0]).toBeInTheDocument();
+    expect(getAllByText('Total sales')[0]).toBeInTheDocument();
+    expect(getAllByText('Average order value')[0]).toBeInTheDocument();
+  });
```
noranda, line 397: Can we test that it has the right style here?

noranda (approved):
Looks good with a minor comment about the test.

packages/recipe/src/components/EzTable/__tests__/EzTable.test.tsx
danidewitt: you have an easy way to do that? are we doing it elsewhere? can we just rely on chromatic visual regression tests for that?

noranda: I would remove this test and add a visual regression test with a different background color. Currently both the table and background are white.

danidewitt: seems like they're not both white here, this test seems to cover it https://ezcater.chromatic.com/component?appId=628e5250d2515a004ac66087&csfId=eztable--tables-with-a-transparent-background&buildNumber=467&k=6421fc6ce362e22452e20d9a-1200-interactive-true&h=8&b=-2

PR #923, opened 2023-03-15
feat: update fulfillment fonts

noranda:
## What did we change?
- Updates Fulfillment fonts to use `Roboto Flex` for default fonts and `Montserrat` for header fonts.
- Since fulfillment doesn't have a separate stitches theme like marketplace does, this changes the default Recipe stitches fonts to be the same as the new fulfillment fonts and then adds styles to the marketplace stitches theme to override these back to their `Lato` font. Since we are phasing out stitches, this was the easiest approach.
- Creates new `defaultFont` and `headerFont` stitches variables to be used with stitches themes and adds them to all components still using stitches. Note: `EzGlobalStyles` is used to provide global stitches styles, however since we don't require its use in downstream apps for the themes to work, this adds the `fontFamily` style to both the global styles and individual components. We do not have this issue with emotion.
- Updates the `Getting Started` and `Theming` doc pages with the new fonts.
- Moves `@recipe-ui/theme-marketplace` into Recipe - this package was already deprecated, but we were still using it. It needed to be moved to prevent a circular dependency issue.
- Updates the v17 migration guide to reference changes to fonts and moving marketplace stitches theme into Recipe.
- Updates Storybook and Chromatic config to include "side-by-side" theme for both marketplace and fulfillment - this involved approving a completely new baseline across all components to make sure all fonts were changed (or not changed) correctly.
- Tweaks some chromatic regression tests to better display interactive elements such as drop downs and modals.
- Adds [Turbosnap](https://www.chromatic.com/docs/turbosnap) to Chromatic to only run snapshots for files that have actually changed, saving us build time and money.

Note: Up until now, we were not using the theme switcher on the Recipe doc site correctly. This PR fixes that, meaning switching the theme will now switch the theme for the doc site as well. Try it out!

## Why are we doing this?
https://ezcater.atlassian.net/browse/DF-554
https://ezcater.atlassian.net/browse/DF-567

### Chromatic example:
Screenshot 2023-03-15 at 3 56 17 PM

### Migration guide:
Screenshot 2023-03-15 at 3 56 44 PM

### Theming page:
Screenshot 2023-03-15 at 3 57 18 PM

## Checklist


pgrippi (approved):

.changeset/rotten-timers-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ezcater/recipe': major
+'@ezcater/recipe-doc-site': major
+---
+
+BREAKING CHANGE: fulfillment theme fonts changed to Montserrat (headers) and Roboto Flex
```
danidewitt, line 6: i could be convinced - is this a breaking change if it doesn't affect APIs (according to this [here](https://github.com/ezcater/recipe/blob/9d26fca8b89a4a8d01b279dceca9e03057c8293a/docs/src/pages/guides/versioning-and-changelog.md?plain=1#L28)). do we need a changeset for the updated deps?

danidewitt: is it because we are recommending `stitchesMarketplaceTheme`?

danidewitt (commented):
left a few minor comments on this. overall nice job on this one! there was a lot of really old stuff that wasn't working that you had to discover and update. and then some other good new stuff you added (like the chromatic theming stuff).

.changeset/rotten-timers-kiss.md
noranda: I was on the fence about this as well, but since we're doing a major release anyways and it certainly will affect the look of every app that has text, I thought it a conservative move. It did also allow me to reiterate that the marketplace theme package is deprecated (even though we were still using it) and should be used directly from Recipe now. I'll add a changeset for the updated deps.

danidewitt (approved):

PR #925, opened 2023-03-16
feat: add filterTerm prop to EzTable

danidewitt:
## What did we change?
- Add `filterTerm` prop to `EzTable`

## Why are we doing this?
- Designer request

## Screenshot(s) / Gif(s):
Screen Shot 2023-03-16 at 10 50 29 AM

## Checklist


paul-ez (dismissed):

packages/recipe/src/components/EzTable/EzTable.md
```diff
@@ -142,6 +142,60 @@ Consider wrapping actions in an [EzLayout](/components/ez-layout) to manage how
 </EzPage>
 ```
 
+### Table with search
+
+To add a global filter search to a table, use the `globalFilter` prop and `EzSearchInput` as a table action.
+
+If the global filter needs to be disabled in the case of very large datasets, the recommendation is to first work with a designer to show usable tables with more user-friendly pre-filtered datasets. If that is not possible and performance becomes an issue, use the `disableGlobalFilter` prop.
+
+If the globabl filter needs to be disabled for backend pagniation purposes, use the `disableGlobalFilter` prop.
```
noranda, line 151: Typo.
```suggestion
If the global filter needs to be disabled for backend pagniation purposes, use the `disableGlobalFilter` prop.
```

noranda (changes requested):
A few things here. Could we change `globalFilter` to `filterTerm`? I think the word "global" is a bit confusing (is there a local filter?) and "term" makes it more clear it's a string that you're filtering by. Also, I think we can get rid of `disableGlobalFilter` entirely and let the users define the logic on their own. For example:

```js
() => {
  const [searchTerm, setSearchTerm] = React.useState('');
  const items = [
    {id: '#004', date: '02/28/18', total: '$3267', name: 'Tacodeli'},
    {id: '#005', date: '02/28/18', total: '$3214', name: 'Torchys'},
    {id: '#006', date: '03/01/18', total: '$6212', name: 'Tacobell'},
    {id: '#007', date: '04/10/18', total: '$2219', name: 'Chipotle'},
    {id: '#008', date: '05/13/18', total: '$2713', name: 'Naco Taco'},
  ];
  const disableFilter = items.length > 5;
  const searchInput = (
    <EzSearchInput
      placeholder="Search"
      aria-label="Search orders"
      value={searchTerm}
      onChange={e => setSearchTerm(e.target.value)}
      disabled={disableFilter}
    />
  );
  return (
    <EzPage>
      <EzTable
        title="All Orders"
        actions={searchInput}
        filterTerm={disableFilter ? null : searchTerm}
        columns={[
          {heading: 'Order number', key: 'id'},
          {heading: 'Event date', key: 'date'},
          {heading: 'Total', key: 'total', numeric: true},
          {heading: 'Name', key: 'name'},
        ]}
        items={items}
      />
    </EzPage>
  );
};
```

Thoughts?

(This also needs to be rebased to `v17`.)

danidewitt:
@noranda please see https://react-table-v7.tanstack.com/docs/api/useGlobalFilter#table-options for why i chose global filter and disable global filter

noranda:
> @noranda please see https://react-table-v7.tanstack.com/docs/api/useGlobalFilter#table-options for why i chose global filter and disable global filter

Ah ok, but do we need to have the same API as React Table? I still think the disable prop is not needed here.

PR #926, opened 2023-03-16
refactor: remove emotion from EzWell

noranda:
## What did we change?
Refactors `EzWell` to use stitches without emotion.

## Why are we doing this?
Using both emotion and stitches together does not play nice with Gatsby and was causing MUI to override theme variables. It was breaking in our doc site which uses `EzWell` for all previews. Removing emotion fixes the issue.

## Screenshot(s) / Gif(s):
Broken:
Screenshot 2023-03-16 at 1 07 22 PM

Fixed:
Screenshot 2023-03-16 at 1 07 05 PM

## Checklist


danidewitt (approved):

packages/recipe/src/components/EzWell/EzWell.tsx
```diff
@@ -1,55 +1,42 @@
...
+/**
+ * A Well is a content container that displays non-editable content separate from other content on the screen.
+ * Often this is used to display preformatted text, such as code/markup examples on a documentation page.
+ */
+const EzWell = forwardRef<Ref, EzWellProps>(({children, role, ...otherProps}, ref) => {
+  useEffect(() => {
+    if (!role && (otherProps['aria-label'] || otherProps['aria-labelledby'])) {
+      warning(false, '*Recipe Warning*. EzWell - A labelled Well must have a role.');
+    }
+  }, [otherProps, role]);
```
pgrippi, line 26: tiny nitpick: passing in `otherProps` like this will cause this hook to refire every time this component re-renders, because the object is a "new" object every time (e.g., it's not memoized)

I think a better solution is to pull the aria label check outside the hook, and pass in that boolean as a dependency to the hook:

```jsx
const hasAriaLabel = Boolean(otherProps['aria-label'] || otherProps['aria-labelledby']);
useEffect(() => {
  if (!role && hasAriaLabel) {
    warning(false, '*Recipe Warning*. EzWell - A labelled Well must have a role.');
  }
}, [role, hasAriaLabel]);
```

pgrippi (approved):

packages/recipe/src/components/EzWell/EzWell.tsx
noranda: I like it!

PR #944, opened 2023-05-05
feat: add onKeyDown to options in EzSegmentedControl

danidewitt:
## What did we change?
- Added `onKeyDown` to `options` prop for `EzSegmentedControl` with relevant documentation and tests

## Why are we doing this?
- Allow better keyboard selection and deselection (accessibility)

## Screenshot(s) / Gif(s):
Screen Shot 2023-05-05 at 3 25 39 PM

## Checklist


packages/recipe/src/components/EzSegmentedControl/EzSegmentedControl.tsx
```diff
@@ -129,6 +130,7 @@ const EzSegmentedControl: React.FC<Props> = ({
           onChange={() => onChange && onChange(option.value)}
           disabled={option.disabled}
           onClick={() => option.onClick && option.onClick()}
+          onKeyDown={() => option.onKeyDown && option.onKeyDown(option.value)}
```
pgrippi, line 133: would it make more sense to pass the event on `onKeyDown`? something like:

```tsx
onKeyDown={event => options.onKeyDown?.(event, value)}
```

I think consumers like store-next are going to way to know what key was pressed, to react accordingly.

packages/recipe/src/components/EzSegmentedControl/EzSegmentedControl.md
```diff
@@ -135,6 +135,68 @@ Used to trigger an `onClick` handler when an option is clicked. Does not conflic
...
+  const OPTIONS_WITH_HANDLERS = OPTIONS.map(option => ({
+    ...option,
+    onClick: handleOnClick,
+    onKeyDown: handleOnKeyDown,
+  }));
+
+  return (
+    <>
+      <EzSegmentedControl
+        name="view-map-options-with-onclick-option"
```
pgrippi, line 188: I think this needs a different name, it's connected to the example above it:

https://user-images.githubusercontent.com/1180841/236555349-4772f06e-83e0-436e-8154-5c7caacf15e1.mp4

pgrippi:
I think ecomm is looking for the ability to use tab and keyboard controls like enter/space to change the options. using `onKeyDown` on the options themselves will allow us to change the value when enter/space is used, but tabbing between the options doesn't work.

@amberlcook after playing around with this component I noticed that the arrow keys are used to change the active value:

https://user-images.githubusercontent.com/1180841/236557534-488b42dc-2c80-4f32-b9fb-3702b9711881.mp4

I think that's why we can't tab between the elements - we're not supposed to!

so with that in mind:

1. since we can change the value using left+right, is that fine for accessibility? I imagine this component was designed with that interface in mind so maybe we were just testing it wrong?
2. if not, maybe having an `isTabbable` option on `EzSegmentedControl` would be a better interface? then we don't need to intercept keydown, it will just change the behavior internally in recipe to be tabable buttons instead of one "button group"

danidewitt:
@pgrippi i already had put that information about the guidance on tab vs. arrows in the documentation and the slack thread

packages/recipe/src/components/EzSegmentedControl/EzSegmentedControl.md
danidewitt: nice catch, i used the previous example as a template

packages/recipe/src/components/EzSegmentedControl/EzSegmentedControl.tsx
danidewitt: hm, i can be convinced, but why would do we need to know which key

pgrippi: well the way it is now, if the user hits any key on the keyboard it'll toggle off the value, I think we only want certain keys to toggle the value

noranda (changes requested):
There are two issues here that I think need to be addressed:

1. The focus outline on an unselected option is not closed. It looks fine for selected options.
Screenshot 2023-05-08 at 9 25 06 AM
Screenshot 2023-05-08 at 9 29 03 AM

2. The keyboard selection behavior works fine for "enter", but not for "space", which will both unselect and then reselect an option.
https://user-images.githubusercontent.com/5418735/236837800-ae1ea9f7-f7c8-4024-bf55-69be4db1bf1e.mp4

danidewitt:
@noranda seems to be a pre-existing styling issue

pgrippi (approved):

amberlcook (approved):
This looks great!! Thanks so much for working on this so quickly.

noranda (approved):

PR #946, opened 2023-05-09
fix: update EzDateInput value when props change

danidewitt:
## What did we change?
- Update `EzDateInput` value when props change

## Why are we doing this?
- Programmatically change value of date input (example: clear button)

## Screenshot(s) / Gif(s):
Screen Shot 2023-05-09 at 12 11 21 PM

## Checklist


pgrippi (approved):

mathmerized (approved):

packages/recipe/src/components/EzField/EzField.md
```diff
@@ -400,6 +400,29 @@ The below example demonstrates using the `filterDate` prop to restrict the calen
...
+        helperText="This is the date your food will be delivered."
+        onChange={value => setDate(value)}
+      />
+      <EzButton onClick={handleClear}>Clear</EzButton>
+    </EzFormLayout>
+  );
+};
+```
+
 Times in this component are displayed with the format h:mma, for example, `9:45am`.
```
noranda, line 426: Was this line supposed to go with the previous example? Or wait, isn't this for times, not dates?

noranda (approved):
Should we add a test for this? Otherwise, looks good!

packages/recipe/src/components/EzField/EzField.md
noranda: Haha, it seems to have been mixed up with `Dates in this component are displayed with the format MM/DD/YYYY.` under the Time Input Field section. I would switch them back!

pgrippi: maybe this line somehow got swapped with line 460? that line seems to fit better here and this line seems to fit better there!

https://github.com/ezcater/recipe/blob/33a2e98ffc77a0efc66b3af5d41a58f62029f90b/packages/recipe/src/components/EzField/EzField.md?plain=1#L460

danidewitt: @noranda as for adding a test, i was trying to avoid testing external local state. is there something else to test here?

noranda: Yeah good point, although this is not something that chromatic can catch easily. It is however the perfect interaction test once we move this to Storybook Docs, so I think it's good for this PR. I'll just have to try to remember to add it later.

PR #953, opened 2023-05-12
docs: add advanced filtering examples to EzTable

danidewitt:
## What did we change?
- Add an example of advanced filtering to EzTable

## Why are we doing this?
- Squad request

## Screenshot(s) / Gif(s):
Screen Shot 2023-05-12 at 3 21 13 PM

## Checklist


packages/recipe/src/components/EzTable/EzTable.md
```diff
@@ -894,3 +894,392 @@ When offering pagination and selection, users should be offered options:
   return <Table />;
 };
 ```
+
+### Advanced filtering
+
+An example of using advanced table filtering.
+
+This is a presentationl example of how to use multiple filters in a table.
```
agrobbin, line 902:
```suggestion
This is a presentational example of how to use multiple filters in a table.
```

packages/recipe/src/components/EzTable/EzTable.md
```diff
@@ -894,3 +894,392 @@ When offering pagination and selection, users should be offered options:
...
+
+To avoid performance degradation
+
+- Do NOT use client-side filtering for large datasets, it can make the table unusable. Prefer server-side filtering, in most cases.
+- [Debounce](https://www.developerway.com/posts/debouncing-in-react) queries as filters change.
```
agrobbin, line 907: I'd also be curious what you think about rephrasing the recommendation sentence and debouncing bullet point:

> Prefer making server-side API calls to filter the actual dataset as the client-side filters change, with appropriate [debouncing](https://www.developerway.com/posts/debouncing-in-react) based on the situation.

It's a subtle change, but I think it clarifies that we don't mean moving all of the client-side logic server-side, just the actual filtering of the dataset.

agrobbin (commented):
Thanks for tagging me here @danidewitt! This is my first real Recipe PR review, so hopefully it's helpful!

packages/recipe/src/components/EzTable/EzTable.md
noranda, line 902: Perhaps we could combine these to simplify?
```suggestion
This is an advanced example for using multiple filters in a table.
```

packages/recipe/src/components/EzTable/EzTable.md
```diff
@@ -894,3 +894,392 @@ When offering pagination and selection, users should be offered options:
...
+
+```jsx
+() => {
+  /* This example code may be too long to run in the playroom, 
+     in order to run in the playroom, pare down the example
+  */
```
noranda, line 913: If this doesn't work in the playroom, can we add a snippet that does and refer to it here? Or can we fix this example so that it works? Having a working playroom example here seems useful.

packages/recipe/src/components/EzTable/EzTable.md
```diff
@@ -894,3 +894,392 @@ When offering pagination and selection, users should be offered options:
...
+    const [startDate, setStartDate] = React.useState('');
+    const [endDate, setEndDate] = React.useState('');
+    const [status, setStatus] = React.useState(null);
+    const [address, setAddress] = React.useState('');
+    const [addressSearchTerm, setAddressSearchTerm] = React.useState('');
+    const [data, setData] = React.useState(allData);
+
+    // This filters table data
+    // This is a presentational example
+    // PLEASE do this in more performant way: server-side debounced filtering logic preferred
```
noranda, line 1025:
```suggestion
    // PLEASE do this in a more performant way: server-side debounced filtering logic preferred
```

noranda (changes requested):
I'm a little worried that this example is so complex that its usefulness is limited. The example itself is almost 400 lines long. Do we need this level of detail to get the point across? Can this be broken into a few smaller examples that would be easier to reason with? This may be useful for a very specific use case right now, but is it useful going forward? When we move to an MUI-based implementation, I'm sure this will get simplified, if you'd rather deal with it then.

Some issues I found with testing:
1. When selecting an address and clicking the Clear button, the address does not get cleared.
2. If I select a date, then Clear, it clears fine, but if I do it again, the Clear button doesn't work.
3. Some of the status options are cut off.
4. Inserting an end date only doesn't seem to work without a start date (but it works the other way around). Is this as designed? Should the user want a range from [any date] to [end date]?

I would also run `yarn install` on the root level on this branch. When I did so, there were changes.

danidewitt:
@noranda 

This example is useful because it was requested by our customers as high priority — both a designer and engineering team. We had a meeting dedicated to it. It’s definitely a use case multiple teams are asking for. So yes, it’s useful now and it’s useful going forward. And we’ll update it when we refactor the tables and/or add a composed components layer, but I’m trying to satisfy an existing need. 

The entire purpose of it is to help engineering teams build complex table filtering without having to figure out it out from scratch every time, on multiple projects. If we don’t add this level of detail, it is useless because the point of it is to show how to use multiple types of filters simultaneously. Breaking it into smaller examples would defeat that purpose.

To address your list of concerns:
1. Clearing the address — this is an existing bug in EzAutosuggest, similar to the bug I had to fix for this ticket in [EzDatePicker](https://github.com/ezcater/recipe/pull/946). I made a JIRA ticket for it.
2. Selecting a date, clearing it, and reselecting a date — I can’t reproduce this issue?
3. Just a layout issue — totally up to the team and designer. I don’t have their mockup for this, just that they wanted to use this combination of filters together.
4. Selecting just an end date — I fixed this. Yes, you’d want to be able to select one date filter or both.
5. What were the yarn install changes? I don’t see any?

packages/recipe/src/components/EzTable/EzTable.md
noranda: Oh yeah, we definitely don't want to do that. I guess I'm okay leaving this as is, so long as we prioritize Tables as the next big redesign. 👍

pgrippi (approved):

PR #957, opened 2023-05-16
Fix storybook doc guide links and update plop CLI tool

noranda:
_Note: This branch is based off `nb/update-emotion-colors` and should therefore be merged after https://github.com/ezcater/recipe/pull/956._

## What did we change?
- [Fix storybook docs guide links and update guides where appropriate](https://github.com/ezcater/recipe/commit/70ec118f4280c3c0d04c1ed9227d1be9b5b3291e)
- [update plop CLI tool to create new storybook docs structure](https://github.com/ezcater/recipe/commit/f0ac96c1b42eabaa5285a6b9e0d9201b54e499b2)

## Why are we doing this?
This is part of an effort to switch our doc site from Gatbsy to Storybook Docs.

PR #960, opened 2023-05-18
Move EzBadge to Storybook Docs

noranda:
_Note: This branch is based off `nb/fix-doc-guide-links` and should therefore be merged after https://github.com/ezcater/recipe/pull/957._

## What did we change?
- [move EzBadge to storybook docs](https://github.com/ezcater/recipe/commit/feab40c68586d065d4f3f17373f5184954530a8f)
- [update language for accessibility color contrast to be stronger](https://github.com/ezcater/recipe/pull/960/commits/d616ec5923e31c13080295cd7d17cd7f4179bfb6)

## Why are we doing this?
This is part of an effort to switch our doc site from Gatbsy to Storybook Docs.

_Note: This is an internal documentation change that doesn't affect the recipe core library and therefore has no changeset._

## Screenshot(s) / Gif(s):
Screenshot 2023-05-18 at 2 12 24 PM
Screenshot 2023-05-18 at 2 12 45 PM
Screenshot 2023-05-24 at 8 56 12 AM
Screenshot 2023-05-18 at 2 13 14 PM
Screenshot 2023-05-18 at 2 13 32 PM


pgrippi (approved):

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+To cap the value of a badge, provide a numerical `max` value. Any value over the `max` will show `max+`. The default is `99`.
+
+<Canvas of={ValueStories.Max} meta={ValueStories} />
+
+## Badge Colors
+
+EzBadge supports theme palette properties (`primary`, `secondary`, `error`, `warning`, `alert`, `info`, `nuetral`, and `success`) as well as all supported [theme colors](http://localhost:8000/guides/theming/#colors) (ex. `common.red100`).
+
+Out of the box, badges have a background color of `common.red110` and font color of `common.white`. Text and background colors should have high enough [contrast](https://webaim.org/resources/contrastchecker/) for accessibility.
```
agrobbin, line 57: This is subtle, but we use "should" here, and I'm wondering if that's an intentional choice. For something like color contrast + accessibility, should (heh) we use "must"?

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+  <Story of={ColorStories.AlertColor} meta={ColorStories} />
+  <Story of={ColorStories.InfoColor} meta={ColorStories} />
+  <Story of={ColorStories.NeutralColor} meta={ColorStories} />
+  <Story of={ColorStories.SuccessColor} meta={ColorStories} />
+  <Story of={ColorStories.CommonColor} meta={ColorStories} />
+</Canvas>
+
+## Aligning Badges
+
+The default alignment for a badge is `top` `right`. To align a badge with its wrapped element, use `alignX` with `left` or `right` and `alignY` with `top` or `bottom`.
```
agrobbin, line 80: Seeing `top` and `right` immediately next to each other looks a little funky. Would it be clearer to spell out the full `alignX=right alignY=top`?

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+To align the badge relative to the corner of the wrapped element, use `overlap` with a value of `circular` or `rectangular`, depending on the shape of the wrapped element. The default is `rectangular`.
+
+<Canvas>
+  <Story of={AlignmentStories.Rectangular} meta={AlignmentStories} />
+  <Story of={AlignmentStories.Circular} meta={AlignmentStories} />
+</Canvas>
+
+## Accessibility
+
+Badges should be wrapped in an accessible element with a proper `aria-label` to be read by screenreaders. The wrapping element will depend on the context with which you use the badge.
```
agrobbin, line 98: I *think* this should be "screen readers":

```suggestion
Badges should be wrapped in an accessible element with a proper `aria-label` to be read by screen readers. The wrapping element will depend on the context with which you use the badge.
```

packages/recipe/src/components/EzBadge/EzBadge.tsx
```diff
@@ -24,7 +24,7 @@ const EzBadge = forwardRef<Ref, EzBadgeProps>(
 EzBadge.defaultProps = {
   alignX: 'right',
   alignY: 'top',
-  color: 'error',
+  color: 'common.red110',
```
agrobbin, line 27: I'm surprised this accepts a string of 'common.red110', how exactly does that work?

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
noranda: I would normally agree with you, except I know we currently have color contrast issues in some of our component variants (design is working on defining a more comprehensive color token palette that will fix these issues). This would also need to be fixed in all our components for consistency. I'm ok with this language for now with perhaps a follow up to make this language stronger once we're 100% compliant.

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
noranda: Works for me!

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
noranda: I think you're right, although I've seen it spelled both ways. I will fix.

packages/recipe/src/components/EzBadge/EzBadge.tsx
noranda: We define a set of common colors in our emotion themes. So long as the color is defined there, it will work here as shown. Since we use an MUI implementation here, it will translate any defined colors in our emotion palette, including common colors.

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
agrobbin: Since it looks like we're providing a recommendation here on usage within applications of `EzBadge`, I'd actually suggest that we have the stronger language, even if we aren't actually compliant everywhere, yet.

packages/recipe/src/components/EzBadge/EzBadge.tsx
agrobbin: Interesting, good to know, thanks @noranda!

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
noranda: I've updated the language for this and all other similar components to use stronger language (in a new commit).

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+<Canvas of={ValueStories.ShowZero} meta={ValueStories} />
+
+To cap the value of a badge, provide a numerical `max` value. Any value over the `max` will show `max+`. The default is `99`.
+
+<Canvas of={ValueStories.Max} meta={ValueStories} />
+
+## Badge Colors
+
+EzBadge supports theme palette properties (`primary`, `secondary`, `error`, `warning`, `alert`, `info`, `nuetral`, and `success`) as well as all supported [theme colors](http://localhost:8000/guides/theming/#colors) (ex. `common.red100`).
```
agrobbin, line 55:
```suggestion
EzBadge supports theme palette properties (`primary`, `secondary`, `error`, `warning`, `alert`, `info`, `neutral`, and `success`) as well as all supported [theme colors](http://localhost:8000/guides/theming/#colors) (ex. `common.red100`).
```

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+## Props
+
+<Controls of={DefaultStories.Default} sort="requiredFirst" />
+
+## Best Practices
+
+- Badges convey information about notifications or events.
+- Badges appear very close to, often overlapping, the element that they are describing.
+- Badges should make it easy to understand the presence of notifications or events, encouraging action to be taken on the element the badge describes.
+- Badges should always be wrapped in an accessible element with a proper `aria-label` to be read by screenreaders. The wrapping element will depend on the context with which you use the badge (see below for examples).
```
agrobbin, line 29:
```suggestion
- Badges should always be wrapped in an accessible element with a proper `aria-label` to be read by screen readers. The wrapping element will depend on the context with which you use the badge (see below for examples).
```

packages/recipe/src/components/EzBadge/Documentation/EzBadge.mdx
```diff
@@ -0,0 +1,114 @@
...
+To align the badge relative to the corner of the wrapped element, use `overlap` with a value of `circular` or `rectangular`, depending on the shape of the wrapped element. The default is `rectangular`.
+
+<Canvas>
+  <Story of={AlignmentStories.Rectangular} meta={AlignmentStories} />
+  <Story of={AlignmentStories.Circular} meta={AlignmentStories} />
+</Canvas>
+
+## Accessibility
+
+Badges should be wrapped in an accessible element with a proper `aria-label` to be read by screen readers. The wrapping element will depend on the context with which you use the badge.
```
agrobbin, line 98: Want to switch this "should" to "must" as well?

packages/recipe/src/components/EzBadge/Documentation/Stories/Accessibility.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export const AriaLabel: Story = {
+  args: {
+    ...Default.args,
+  },
+  parameters: {
+    docs: {
+      source: {
+        code: dedent`
+          <div aria-label="3 items in cart">
```
agrobbin, line 31: This isn't something that should block this PR, but I'm curious why `EzBadge` puts this requirement on the user of the component, rather than including it within the scope of responsibility of the component itself.

agrobbin (commented):
This is the first PR that I think I've seen that actually moves a component to Storybook Docs, so I'm not nearly as up-to-speed on the basics of what's required to do so (hence my lack of official Approval).

I'm also a little concerned at the work (and sheer amount of code) required to make this work. The overhead of writing the documentation for a new component now appears significantly more than it was before, and if we want to encourage those outside of Recipe to contribute, we've added a large hurdle to that goal (even with the generator included in #957). Is that "the cost of doing business" these days, or are there opportunities available to us to reduce the effort required?

I'm also wondering if there is a check of some kind we can add to CI that ensures that changes to a component (like adding a new option) are accompanied by a corresponding change to the docs. If we can't automate it, a starting point could be adding a checklist item to our [PR template](https://github.com/ezcater/recipe/blob/da0d5cf50ef1e0e19b527b47c11d7e560c8ed8a9/.github/pull_request_template.md).

packages/recipe/src/components/EzBadge/Documentation/Stories/Accessibility.stories.tsx
noranda: We leave it up to the user because it can be used as a standalone badge with a sibling element (see [example](https://github.com/ezcater/recipe/pull/960/files#diff-988ada539c1873722b806b01b7d5ed6d0c3652a3accc87ddea28be0804efab09R74-R81)). In these cases, both elements need to be wrapped and given an aria-label, which can't be done from this component.

agrobbin (approved):
@noranda and I discussed live the Storybook Docs effort, which helped add to my understanding of what's involved here. Since Storybook Docs is not a new technology to other teams, and we have the generator, while the effort isn't nothing, the value far outweighs whatever it is!

With that in mind, I'm 👍🏻! If anything we missed that is caught when @danidewitt is back, we can definitely do a follow-up.

PR #986, opened 2023-06-29
Make EzSearchInput `type="search"` to make clearable by default

noranda:
## What did we change?
- [make EzSearchInput type search to make clearable by default](https://github.com/ezcater/recipe/commit/9865122004bffe38775565514335cf0e3629d8ce)

## Why are we doing this?

Design [request](https://www.figma.com/file/3T0PuTpgMT2pE5HnbLE4Xp/Corporate-Dashboard?node-id=4665%3A12013&mode=dev) - this PR adds the clearable functionality by default - style of icon cannot be easily changed, so confirmed with design that this will do until this component is redesigned.

## Screenshot(s) / Gif(s):
Screenshot 2023-06-29 at 4 45 34 PM

## Checklist

### DIFF
--- a/packages/recipe/src/components/EzSearchInput/EzSearchInput.tsx
+++ b/packages/recipe/src/components/EzSearchInput/EzSearchInput.tsx
@@ -9,12 +9,13 @@ const searchInput = theme.css({
   '&&': {
     paddingLeft: '$search-input-padding-left',
   },
+  WebkitAppearance: 'none',
 });
 
 type Props = React.InputHTMLAttributes<HTMLInputElement>;
 
 const EzSearchInput = forwardRef<HTMLInputElement, Props>((props, ref) => (
-  <EzTextInput ref={ref} {...props} className={searchInput()} />
+  <EzTextInput ref={ref} {...props} className={searchInput()} type="search" />
 ));
 
 export default EzSearchInput;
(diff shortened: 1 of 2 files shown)


packages/recipe/src/components/EzSearchInput/EzSearchInput.tsx
```diff
@@ -14,7 +14,7 @@ const searchInput = theme.css({
 type Props = React.InputHTMLAttributes<HTMLInputElement>;
 
 const EzSearchInput = forwardRef<HTMLInputElement, Props>((props, ref) => (
-  <EzTextInput ref={ref} {...props} className={searchInput()} />
+  <EzTextInput ref={ref} {...props} className={searchInput()} type="search" />
```
danidewitt, line 17: woah - really nice find for this solution! 🥇

danidewitt (dismissed):

packages/recipe/src/components/EzSearchInput/EzSearchInput.tsx
danidewitt: just wanted to follow up and say it's awesome that this simultaneously: 
1. delivers customer value, and quickly
2. fixes the issue of the input being the wrong type (which i wasn't aware of until now)

noranda:
Note: I ran into an issue with chromatic not rendering the search icon in Safari (due to a version of playright they use for snapshots not rendering it correctly). It renders just fine in Safari directly, but in order to fix this in chromatic, I've added a style to the input.

pgrippi (approved):

danidewitt (approved):

PR #995, opened 2023-07-11
feat: add maxDependentValue and maxIndependentValue to EzLineChart and EzBarChart

danidewitt:
## What did we change?
- Added maxDependentValue and maxIndependentValue to EzLineChart and EzBarChart 

## Why are we doing this?
When a chart has a value above the top axes label, show the next increment on the axes, see below
Screen Shot 2023-06-22 at 3 12 26 PM

## Screenshot(s) / Gif(s):
Screen Shot 2023-07-11 at 1 37 15 PM
Screen Shot 2023-07-11 at 1 37 29 PM

## Checklist


packages/recipe/src/components/EzBarChart/EzBarChart.md
```diff
@@ -77,6 +79,13 @@ In some cases, like server-side rendering, a deterministic id may be needed. If
 
   const DEPENDENT_AXIS_LABEL_VALUES = [0, 50000, 100000, 150000, 200000, 250000, 300000];
 
+  /* start calculation for custom max y-value */
+  const dependentValues = DATA.map(datum => datum.y).sort();
+  const highestDependentValue = dependentValues[dependentValues.length - 1];
```
noranda, line 84: Should we use `Math.max(DATA.map(datum => datum.y))` for this instead?

packages/recipe/src/components/EzBarChart/__tests__/EzBarChart.test.tsx
```diff
@@ -47,7 +55,32 @@ describe('EzBarChart logic', () => {
...
+        maxDependentValue={MAX_DEPENDENT_VALUE}
+        maxIndependentValue={MAX_INDEPENDENT_VALUE}
+        title={TITLE}
+      />
+    );
+
+    expect(getByText(`${MAX_DEPENDENT_VALUE}`)).toBeInTheDocument();
+    expect(getByText(`${MAX_INDEPENDENT_VALUE}`)).toBeInTheDocument();
+    expect(queryByText(`${MAX_DEPENDENT_VALUE + INCREMENT}`)).toBeNull();
+    expect(queryByText(`${MAX_INDEPENDENT_VALUE + 1}`)).toBeNull();
```
noranda, line 83: I think these should be `.not.toBeInTheDocument()` instead of `.toBeNull()` since this is a DOM assertion.

noranda (commented):
How should this work if the `maxDependentValue` is not calculated to be the next highest increment. For example, using the example in the docs, if I make it `6000` or `7000` (see screenshot below), it will display the `5000` divider, but will just display empty space for anything above. Is this by design?

Otherwise, just a few minor comments and this looks good!

Screenshot 2023-07-12 at 3 57 19 PM

danidewitt:
@noranda to answer your question, i wouldn't explicitly define `dependentAxisLabelValues` simultaneously with `maxDependentValue`. if you stop passing that `dependentAxisLabelValues` prop, the axis labels are fine.

packages/recipe/src/components/EzBarChart/EzBarChart.md
danidewitt: yeah, that's quite a bit better 😆

noranda:
> @noranda to answer your question, i wouldn't explicitly define `dependentAxisLabelValues` simultaneously with `maxDependentValue`. if you stop passing that `dependentAxisLabelValues` prop, the axis labels are fine.

Ah I see. It actually seems fine to define both so long as the max is included in the values. I'm trying to think of use cases where we may not know the max of the data, but we know the max possible value. Perhaps we could make this a little more clear in the docs?

noranda (approved):

pgrippi (approved):

PR #996, opened 2023-07-12
fix: remove inline styles using useMemo

danidewitt:
## What did we change?
- Removed instances of icons with inline styling using useMemo

## Why are we doing this?
- Potential render lifecycle issues, in this case reported in downstream tests

## Screenshot(s) / Gif(s):
- see Chromatic tests in CI

## Checklist

- [x] Provide test coverage for the changes made - see Chromatic tests in CI
 

pgrippi (dismissed):
that's really weird if this is causing lifecycle hook errors, none of these `useMemo` calls are conditional / in a loop

noranda:
Can you explain a little more how using `useMemo` in these ways can cause issues? I don't think we're using it incorrectly here (breaking hook rules). I'm not opposed to using MUI instead, I was just curious about the intent here.

noranda (approved):
Approving this with the caveat that while it's not a necessary change, it doesn't hurt and lets us resolve this PR.

pgrippi (approved):

PR #1005, opened 2023-08-04
Convert EzAlert styles from Stitches to MUI+emotion

pgrippi:
## What did we change?

This PR accomplishes two things:

* It lays the foundation for using MUIs style engine to style custom recipe components that are not built on top of MUI
* It refactors `EzAlert` to use MUI instead of stitches for styling

## Why are we doing this?

This lays the foundation needed to remove stitches ([FEC-728](https://ezcater.atlassian.net/browse/FEC-728)) by allowing us to move to MUI for styling our custom components

There is likely more work to be done here, as the `EzAlert` component was fairly straightforward and didn't require styling changes between the marketplace and fulfillment themes.

## Checklist

[FEC-728]: https://ezcater.atlassian.net/browse/FEC-728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/recipe/src/themes/ezColors.ts
```diff
@@ -190,6 +190,53 @@ export const ezColors = {
   },
 } as const;
 
+export const legacyColors = {
```
pgrippi, line 193: Because EzAlert uses the legacy colors I opted to move these colors to this file where I could import it in the theme without creating a circular dependency

packages/recipe/src/themes/mui/ezTheme.ts
```diff
@@ -285,6 +286,82 @@ const components = {
       },
     },
   },
+
+  // Recipe components
+  EzAlert: {
+    variants: [
```
pgrippi, line 292: This is where the magic happen - there are two sets of styles we create in this approach: in EzAlert we have the base container styles (`EzAlertContainer`), and here we have the variants.

The variants have two components, a `props` key that MUI will reference to see which variant(s) match (e.g., which props passed to the components are defined here), and a `style` key whose styles will be applied when the props match!

pgrippi:
This was a wild ride but I pushed 1c1c4b611107200c12ece95818121f9f20232a8f which I think will fix the storybook problem...

What this really boiled down to was me adding the emotion babel plugin to recipe. I added this plugin because it adds some build time niceties that makes local development better (like adding inline sourcemaps), and in theory makes the production code faster as emotion can do some of the work at build time.

This worked fine locally and I was able to inspect both the cjs and esm builds and see that EzAlert compiles to what I would expect with the plugin.

However, this seems to have broken something inside of Storybook (they also use emotion). What I _think_ is breaking is using the `css={}` property, for instance here in EzCard (one of the stories that was failing): https://github.com/ezcater/recipe/blob/98a732cf94c485f434851bbba2daa04458315e72/packages/recipe/src/components/EzCard/EzCardHeading.tsx#L38

Both emotion and stitches add a special `css` property to JSX components, and I think they behave differently. By using the emotion plugin _all_ components use the emotion JSX wrapper, so the stitches-specific stuff will break (which perfectly explains why the card saw a breaking change).

So my solution is a temporary one, I think the path forward looks like:

* Use the emotion babel plugin when building recipe (e.g., `yarn build`, `yarn develop`)
* Do not use the emotion babel plugin when building storybook (`yarn storybook`, `yarn build-storybook`)
* (Both of those points ☝🏻 are introduced in this PR)
* Continue refactoring components to use emotion instead of stitches (eventually refactoring you're one use of `css={}` in EzCard/EzLayout)
* Once stitches has been removed, we can remove the conditional babel code and just use the emotion plugin always (changing the file back to a simple `.babelrc` file)

pgrippi:
@noranda if you get a chance can you look at the chromatic failures here? https://ezcater.chromatic.com/build?appId=628e5250d2515a004ac66087&number=875

I fixed the layout issue but was surprised to see failures, but honestly they all look like they're around the placeholder images being slightly different, just want to get your take on it

noranda (approved):
Thanks for the contribution, Peter! This will be a great help in planning for the complete removal of stitches!

danidewitt (approved):
thank you!

PR #1007, opened 2023-08-08
Add components-only Storybook to Gatsby

danidewitt:
## What did we change?
- Add bare bones, components only, as-is Storybook to Gatsby docs, avoiding delays from creating new repos or engaging other vendors for hosting

## Why are we doing this?
- Create a *temporary* public link to our components in Storybook to test story.to.design and other Figma integrations to unblock design team immediately (as requested by Joel and Mark)

## Screenshot(s) / Gif(s):
- Run `yarn build` at the project root
- Then `yarn run gatsby serve --prefix-paths` in the `docs` directory
- Test at `localhost:9000/recipe/storybook`

Screenshot 2023-08-07 at 8 49 32 PM

## Checklist


noranda:
@danidewitt I don't think this PR is necessary for a few reasons:

1. It doesn't solve the problem. In order to have story.to.design working with our Storybook stories, we need to use Storybook args, as this is what the plugin relies on to extract variants of components (more details [here](https://protect-us.mimecast.com/s/2N8LC4xPV0UB1o59SOPXR1?domain=story.to.design)). When @alexanderlippner16 and I met with story.to.design back in May (I believe you were out), we confirmed this. The Gatsby storybook doesn't include args, so Alexis won't be able to use this to create components in Figma.

2. We already have a published storybook link (found on our #recipe channel bookmarks) through Chromatic that uses the new stories with args. This link currently includes about 1/3 of our components, so we need to finish the conversion before all components could be added, but it gives Alexis a start.

3. We are planning to deprecate the Gatsby doc site in favor of the storybook doc site in the next sprint. As such, it probably makes more sense to focus on this transition (getting all open PRs merged or closed and preparing for `v18`) so that we can move forward.

Feel free to reach out if you want to discuss this further!

noranda (dismissed):
I'm going to approve this PR with the caveat that this alone will not solve the problem we have with needing args. This will also be removed in `v18` but will remain in our `v17` snapshot branch.

pgrippi (dismissed):
one comment about moving `copyfiles` out of dependencies, otherwise lgtm!

We _might_ need to change the webpack public path to `/recipe/storybook` for the doc release, I'm not sure how links will work OOTB because I believe storybook expects to be at the `/` path, but that will be something we will notice once this goes out!

packages/recipe/package.json
```diff
@@ -34,6 +34,7 @@
     "@react-types/provider": "^3.1.1",
     "@react-types/shared": "^3.7.0",
     "@stitches/core": "^1.2.8",
+    "copyfiles": "^2.4.1",
```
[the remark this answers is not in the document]
danidewitt: good catch. i def ran that with the dev dependency flag but somehow it ended up in the wrong place

danidewitt:
@noranda yeah me too. i'm looking into it

danidewitt:
@pgrippi @noranda this is ready for re-review. i ended up having to pin the storybook packages to the versions they were on. this is ok because the storybook feature branch work is current storybook. this PR is only meant to be a short-term solution.

pgrippi (dismissed):

noranda (approved):
I'm not sure why the version of storybook had to be locked here, but as you said, this will not be on `main` once we release `v18`, so all good.

packages/recipe/package.json
```diff
@@ -99,7 +99,7 @@
     "build:esm": "cross-env BABEL_ENV=esm babel src --out-dir dist/esm --extensions \".ts,.tsx\" --source-maps inline",
     "build:types": "tsc -p ./tsconfig.build.json",
     "build-publish": "copyfiles -u 1 \"storybook-static/**/*\" ../../docs/static/storybook",
-    "build": "run-p -l \"build:**\" && yarn run build-publish",
+    "build": "run-p -l \"build:**\" && yarn build-storybook && yarn run build-publish",
```
pgrippi, line 102: if you wanted to simply this a bit I think you can swap `run-p` for `run-s` and change this to simply:

```json
"build": "run-s -l \"build:*\" build-storybook build-publish"
```

pgrippi (approved):

packages/recipe/package.json
danidewitt: oh thats nice, updated, thanks!

pgrippi (approved):

noranda (approved):

PR #1019, opened 2023-08-23
Fix playroom in vercel

pgrippi:
## What did we change?

This updates the Storybook playroom config to point to the correct location, and updates the `build-storybook-playroom` script to only build playroom+storybook

## Why are we doing this?

We noticed that playroom was broken in vercel, and it's because we're now hosting it on a root domain. Previously, using github pages, the recipe docs site was hosted on a `/recipe/` path prefix, so we hardcoded the playroom URL to account for that. Now that we're hosting directly on `recipe.ezcater.com` we need to remove that

We also need to build playroom before storybook, otherwise the `playroom` folder isn't present in the public directory and thus never gets created/uploaded.

The trick here is that `main` doesn't have this `build-storybook-playroom` script defined. I think I was able to work around that, where it first tires to use `build-storybook-playroom` and falls back to just `build-playroom` (which does exist in main), but I have no idea if it'll work so I guess we'll see!

## Checklist


pgrippi:
That preview environment is pretty cool! Confirmed that playroom now works:

Screenshot 2023-08-23 at 10 34 46 AM

noranda (approved):

PR #1027, opened 2023-08-25
Disable the emotion babel plugin until we remove stitches

pgrippi:
## What did we change?

This temporarily comments out the `@emotion/babel-plugin` from the recipe babel config

## Why are we doing this?

I added this babel plugin when I rewrote EzAlert to use MUI+emotion (https://github.com/ezcater/recipe/pull/1005). This plugin enables some build-time optimizations for emotion (as well as enabling component css nesting), but it also rewrites all `css={}` props at build time. Both emotion and stitches use this special css prop, and the emotion plugin was accidentally overwriting the stitches calls and passing in invalid objects for stitches.

There is no way for it to be able to tell what is stitches and what is emotion, so the simplest fix is to just remove it for now. I chose to leave it commented out rather than removing it so we can easily enable it once we're done migrating fully away from stitches

## Screenshot(s) / Gif(s):

Before (notice on line 35 we're passing in the emotion object created on line 11 to the `css` prop):

Screenshot 2023-08-25 at 11 52 34 AM

After (notice that on line 22-25 we're now passing in a plain JS object for the `css` prop):

Screenshot 2023-08-25 at 11 52 54 AM

## Checklist

### DIFF
--- a/.changeset/spotty-emus-deny.md
+++ b/.changeset/spotty-emus-deny.md
@@ -1,0 +1,5 @@
+---
+'@ezcater/recipe': patch
+---
+
+fix: prevent the emotion plugin from rewriting stitches css prop
--- a/packages/recipe/babel.config.js
+++ b/packages/recipe/babel.config.js
@@ -1,6 +1,6 @@
-const isStorybookCommand = () =>
-  process.env.npm_lifecycle_event === 'storybook' ||
-  process.env.npm_lifecycle_event === 'build-storybook';
+// const isStorybookCommand = () =>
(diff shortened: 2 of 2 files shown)


pgrippi:
Also you may wonder "why did we miss this in Storybook?" -> it's because I had already disabled the plugin for the storybook build, as it interferes with other CSS. I should have realized this would probably affect the production build, I'm sorry I didn't test it better the first time!

noranda (approved):

kristenlk (approved):

PR #1029, opened 2023-08-28
Make Recipe compatible with React 18 types [FEC-805]

noranda:
## What did we change?
This PR upgrades Recipe to be compliant with React 18 [types](https://www.npmjs.com/package/@types/react) (specifically with `18.2.21`). These changes are also compliant with our current version of React 17 types and so I did not upgrade us to 18 yet as we are not yet on React 18.

## Why are we doing this?
Apps that use Recipe and React 18 are running into issues when upgrading to React 18 types. While they can still use React 17 types, we'd like to support React 18 types so they can too.

## Testing
To test, upgrade to the latest React [types](https://www.npmjs.com/package/@types/react) (`18.2.21`) and run `yarn type-check`. There should be no errors using either React 17 or React 18 types.

## Checklist

### DIFF
--- a/.changeset/three-donuts-eat.md
+++ b/.changeset/three-donuts-eat.md
@@ -1,0 +1,5 @@
+---
+'@ezcater/recipe': minor
+---
+
+feat: support React 18 types
--- a/packages/recipe/src/components/EzAppLayout/EzAppLayout.tsx
+++ b/packages/recipe/src/components/EzAppLayout/EzAppLayout.tsx
@@ -1,4 +1,4 @@
-import React, {createContext, useContext} from 'react';
+import React, {FC, createContext, type PropsWithChildren, useContext} from 'react';
 import {VariantProps} from '@stitches/core';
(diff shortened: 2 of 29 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "@typescript-eslint/eslint-plugin": "^5.18.0",
     "@typescript-eslint/parser": "^5.18.0",
     "prettier": "^2.8.4",
-    "turbo": "^1.2.5",
+    "turbo": "^1.10.13",
```
pgrippi, line 11: 🎱 did you mean to bump turbo? I'm glad you did, we are very far behind on it, but it might be good to do on its own separate from this!

noranda: Oh oops, yep! I'll move this to a different PR.

danidewitt (approved):

stellabonnie (commented):
Wow this was so fast! Thank you 🙏🎊

PR #1042, opened 2023-09-06
Turn on Chromatic Turbosnap [FEC-854]

noranda:
## What did we change?
- [turn on chromatic turbosnap](https://github.com/ezcater/recipe/commit/7fc8f705868b47fd23647bbcf408ed8c2ea91232)

## Why are we doing this?
We had turned off [turbosnap](https://github.com/ezcater/recipe/pull/981) which was not working correctly with our gatsby site. Now that we have switched to storybook docs, we can turn it back on to help save us snapshots.

_No changeset is needed for this change that doesn't affect the Recipe library._

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -24,6 +24,6 @@ jobs:
           projectToken: ${{secrets.CHROMATIC_PROJECT_TOKEN}}
           workingDir: packages/recipe
           exitOnceUploaded: true
-          onlyChanged: false # TODO: turn on after storybook-docs feature merged
+          onlyChanged: true # turbosnap (see https://www.chromatic.com/docs/turbosnap)
         env:
           STORYBOOK_INCLUDE_REGRESSION_STORIES: 1
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -18,12 +18,26 @@ jobs:
         run: yarn build-playroom
         working-directory: ./packages/recipe
 
+      # this only applies to non-main branches (e.g., feature branches)
       - name: Publish to Chromatic
+        if: github.ref != 'refs/heads/main'
         uses: chromaui/action@v1
         with:
           projectToken: ${{secrets.CHROMATIC_PROJECT_TOKEN}}
           workingDir: packages/recipe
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -24,6 +24,6 @@ jobs:
           projectToken: ${{secrets.CHROMATIC_PROJECT_TOKEN}}
           workingDir: packages/recipe
           exitOnceUploaded: true
-          onlyChanged: false # TODO: turn on after storybook-docs feature merged
+          onlyChanged: true # turbosnap (see https://www.chromatic.com/docs/turbosnap)
```
pgrippi, line 27: 🎱 how would you feel about splitting this chromatic action in two: one for the `main` branch and one for all other feature branches. We did this in store-next and I think it could be valuable to do it here too:

```yaml
      # this only applies to non-main branches (e.g., feature branches)
      - name: Publish to Chromatic
        if: github.ref != 'refs/heads/main'
        uses: chromaui/action@v1
        with:
          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
          workingDir: packages/recipe
          exitOnceUploaded: true
          onlyChanged: true

      # this only applies to the `main` branch
      - name: Publish to Chromatic
        if: github.ref == 'refs/heads/main'
        uses: chromaui/action@v1
        with:
          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
          workingDir: packages/recipe
          exitOnceUploaded: true

          # the two main things to point out as being different:
          #   1. onlyChanged is omitted, so `main` will always run the full test suite
          #   2. autoAcceptChanges is set to true, so `main` will always be considered the gold standard
          autoAcceptChanges: true
```

I think setting `autoAcceptChanges` on main should alleviate this problem, where changes are made in a feature branch, accepted as the new baseline, but then the merge into main is seen as different and you need to go into it and manually reapprove it:

Screenshot 2023-09-06 at 1 52 48 PM

noranda: I like it!

pgrippi (approved):

danidewitt (approved):

PR #1066, opened 2023-10-12
fix: add accessible title to truncated EzCard headings

danidewitt:
## What did we change?
- Added accessible title to truncated heading for EzCard

## Why are we doing this?
- Accessibility

## Screenshot(s) / Gif(s):
Screenshot 2023-10-12 at 10 32 18 AM

## Checklist


packages/recipe/src/components/EzCard/EzCardHeading.tsx
```diff
@@ -46,10 +46,15 @@ const EzCardHeading: FC<HeadingProps> = ({
...
+  const accessibileTitle = `
+    ${title}
+    ${subtitle}
+  `;
+
   const heading = title && (
     <EzLayout layout="basic" alignY="top">
       {titleIcon}
-      <EzHeading size="3" subheading={subtitle}>
+      <EzHeading size="3" subheading={subtitle} title={truncateHeading ? accessibileTitle : null}>
```
noranda, line 57: A minor note, but best practice is to use `undefined` here which means "do not use this attribute". See [definitions](https://exploringjs.com/impatient-js/ch_undefined-null.html#undefined-vs.-null).

```suggestion
      <EzHeading size="3" subheading={subtitle} title={truncateHeading ? accessibileTitle : undefined}>
```

noranda (changes requested):
Can the titles be split here? So hovering over the title will just show the title, and hovering over the subtitle will just show the subtitle. It seems odd to combine them like this for both.

noranda (approved):
🚀

jmargolisvt (approved):

pgrippi (approved):

PR #1067, opened 2023-10-13
Fix various accessibility issues in Recipe [FEC-956]

noranda:
## What did we change?
- [fix: add prefix to unique ids for react 18 to fix aria labels which must start with a letter](https://github.com/ezcater/recipe/commit/492c493e93dcd4cdf4eb0d050053b5b9b0f2b27b)
- [fix: for EzFlashMessage components with no headline, apply label `id` to message instead for accessibility](https://github.com/ezcater/recipe/commit/5326021a5574152abc1ba7e51ef30de6f945c01a)
- [fix: add unique ids to EzBarChart and EzLineChart axis labels for accessibility](https://github.com/ezcater/recipe/commit/737f2760cf83e4e34cbcbba0e60acd2bde95b4c8)
- [fix: add role to dropdown input elements when used with `aria-expanded` for accessibility](https://github.com/ezcater/recipe/commit/80ef6d6a1ab19d62ca9ac4b3691b576ba4ee1800)
- [fix: add `aria-hidden="true"` and `focusable="false"` attributes to decorative icons for accessibility](https://github.com/ezcater/recipe/commit/16301b8107e2fba221a16385a268092aa2683912)
- [fix: add `aria-controls` for all inputs with `role="combobox"` regardless if input is open for accessibility](https://github.com/ezcater/recipe/commit/2fd888eb24002e888b9d5a24cad3f738b4507b7c)
- [fix: add `role="presentation"` to line items table in EzOrderSummary for accessibility](https://github.com/ezcater/recipe/commit/9fcc963d72bf4b88aa06555bc6a7d0cff4d9f927)
- [fix: add `role="table"` to EzTable to distinguish from `role="presentation"` for accessibility](https://github.com/ezcater/recipe/commit/f4ccd292351271c27a410e8a62eb2c707069ebe8)
- [fix: add `role="meter"` to EzProgress and make accessible](https://github.com/ezcater/recipe/commit/b4874aaff459aae07192a775883f69291a3bbb69)
- [deps: upgrade `jest-axe` from `3.2.0` to `8.0.0` and fix all tests](https://github.com/ezcater/recipe/commit/b50c0e9b2f6af371ca358c7f4a42f08daa130043)
- [fix: change EzProgress hidden svg to `presentation` role for accessibility](https://github.com/ezcater/recipe/pull/1067/commits/4c9a6f8adc2806f508690dfab9540971e322123f) 

## Why are we doing this?
In general, we want Recipe to be as accessible as possible. This is the first step of more to come. Specifically, this is in response to a recent accessibility [scan](https://docs.google.com/spreadsheets/d/1LJLbK36gGI2s3ktyrF3VIMmLFi8kuJp1u0FwQxNHVJE/edit#gid=1258881448).

## Checklist


packages/recipe/package.json
```diff
@@ -96,7 +96,7 @@
     "esbuild-loader": "^3.0.1",
     "eslint-plugin-storybook": "^0.6.12",
     "jest": "^28.1.2",
-    "jest-axe": "^3.2.0",
+    "jest-axe": "^8.0.0",
```
danidewitt, line 99: i assume there are no breaking changes here since it's jest

danidewitt (approved):

packages/recipe/package.json
noranda: Correct! This was needed for axe to support the newish `meter` role used in EzProgress.

packages/recipe/src/components/Overlays.ts
```diff
@@ -42,7 +42,8 @@ const useOverlayTrigger = (state: OverlayTriggerState) => {
   return {
     triggerProps: {
       'aria-expanded': isOpen,
-      'aria-owns': isOpen ? overlayId : null,
+      'aria-controls': overlayId,
+      role: 'combobox',
```
pgrippi, line 46: ❓ `useMenuTrigger` (which uses this `useOverlayTrigger`) is used by `EzDateInput` and `EzNavigation/UserMenu`, is `role="combobox"` + `aria-controls` right for them too?

packages/recipe/src/components/EzProgress/Implementations/EzProgressMui/EzProgressMui.tsx
```diff
@@ -3,40 +3,46 @@ import Stack from '@mui/material/Stack';
...
 const CIRCULAR_PROGRESS_THICKNESS = 2;
 const CIRCULAR_PROGRESS_FULL_VALUE = 100;
 
-const EzProgressMui: React.FC<EzProgressMuiProps> = ({value, color, children}) => {
+const EzProgressMui: React.FC<EzProgressMuiProps> = ({children, color, label, value}) => {
   return (
     <Stack alignItems="center" justifyContent="center" p="20% 0" position="relative">
       <CircularProgress
-        variant="determinate"
+        aria-hidden="true"
```
pgrippi, line 20: 🎱 should we also unset `role` (e.g., `role={undefined}`) for this `CircularProgress`? though doing that probably gets weird too because the MUI component sets the `valuenow` aria attribute automatically, and then it wouldn't have role to go with that attribute

I know we're hiding it so the role doesn't really matter, but we're also not using it as a progressbar, the same way we're not using the meter below as a progressbar

🤷🏻 `role="presentational"`? 😂

pgrippi (approved):

packages/recipe/src/components/Overlays.ts
noranda: Yes, based on the difference between the two (see [article](https://accessibleweb.com/question-answer/whats-the-difference-between-aria-controls-and-aria-owns/#:~:text=Answer,controlled%20by%20the%20current%20element.)). `aria-controls` is used to note when an element controls another element (for example an element that opens another element), whereas `aria-owns` is just meant to note when a child element belongs to a parent element.

packages/recipe/src/components/EzProgress/Implementations/EzProgressMui/EzProgressMui.tsx
noranda: I think this makes sense. I'll change to `role="presentation"`.

pgrippi (approved):

danidewitt (approved):

PR #1075, opened 2023-10-25
[FEC-996] Remove unused `sosia-markdown` package

noranda:
## What did we change?
This PR removes the `sosia-markdown` package which is no longer in use.

## Why are we doing this?
We are [archiving](https://github.com/ezcater/sosia/pull/116) the `sosia` repo as it is no longer in use and contains security vulnerabilities. We stopped using `sosia` in `16.0.0` when switching to Chromatic and stopped using one of its packages `sosia-markdown` in `18.0.0` when we switched to Storybook Docs.

This is an internal cleanup PR and does not require a changeset.

### DIFF
--- a/packages/recipe/package.json
+++ b/packages/recipe/package.json
@@ -108,7 +108,6 @@
     "react-dom": "^17.0.2",
     "react-markdown": "^8.0.6",
     "react-router-dom": "^5.1.2",
-    "sosia-markdown": "^1.1.2",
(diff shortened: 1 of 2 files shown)


jayroh (approved):
✂️ ✂️ ✂️ ✂️ ✂️ ✂️

pgrippi (approved):

PR #1076, opened 2023-10-25
feat: Add EzStackedBarChart component

danidewitt:
## What did we change?
- Add `EzStackedBarChart` component

## Why are we doing this?
- Design request

## Screenshot(s) / Gif(s):
Screenshot 2023-11-01 at 6 32 30 PM
Screenshot 2023-11-01 at 6 32 36 PM

## Followup v2:
- [ ] Legend squares should have a selection state when the corresponding bar is hovered (design says this can wait until v2 if needed)
- [ ] Grey highlight behind hovered stacked bar (design says this can wait until v2 if needed)
- [x] Abstract custom tooltip more - allow any html inside it as children and also have the current default, remove inline styles if possible in svgs
- [x] Any new accessibility we have added to the other charts (will be done in separate ticket after victory upgrade)

## Checklist


packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
```diff
@@ -0,0 +1,109 @@
+import {Canvas, Controls, Meta, Primary, Source, Story, Unstyled} from '@storybook/blocks';
```
noranda, line 1:
```suggestion
import {Canvas, Controls, Meta, Primary, Source, Story} from '@storybook/blocks';
```

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
```diff
@@ -0,0 +1,109 @@
...
+import dedent from 'ts-dedent';
+import TableOfContents from '../../../../docs/components/TableOfContents';
+import RelatedComponents from '../../../../docs/components/RelatedComponents';
+import EzAlert from '../../EzAlert';
```
noranda, line 5:
```suggestion

```

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
```diff
@@ -0,0 +1,109 @@
...
+<Primary />
+
+## Props
+
+<Controls of={DefaultStories.Default} sort="requiredFirst" />
+
+## Best practices
+
+- Bar charts are used to show change over time, especially over multiple categories.
+- Bar charts are good for comparing relative size.
```
noranda, line 28: Can we add some verbiage here to make clear that this component is only to be used when there is more than one data set? And link to `EzBarChart` if not?

packages/recipe/src/components/EzStackedBarChart/Documentation/Stories/Default.stories.tsx
```diff
@@ -0,0 +1,253 @@
...
+              title="Total spend over the last 12 months."
+            />
+          </EzCard>
+        );
+      })()}
+      `,
+    },
+  },
+  render: args => (
+    <Box sx={{width: '50%'}}>
```
noranda, line 243: Could we make this a little wider? The tooltips are a little tough to read at this width.

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
```diff
@@ -0,0 +1,109 @@
...
+Colors must have high enough [contrast](https://webaim.org/resources/contrastchecker/) for accessibility.
+
+<Source
+  code={dedent`
+    type Datum = {
+      x: string | number | string[] | (() => void);
+      y: string | number | string[] | (() => void);
+    };
+
+    export type DisplayData = {
```
noranda, line 45:
```suggestion
    type DisplayData = {
```

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.snapshot.tsx
noranda, line 1: As this is only a preview, I would recommend not rendering the component here and instead rendering something that looks like the component (see `EzBarChart`), otherwise you can run into rendering issues when it's clicked on.

packages/recipe/src/components/EzStackedBarChart/Documentation/Stories/Default.stories.tsx
```diff
@@ -0,0 +1,253 @@
...
+import {type Meta, type StoryObj} from '@storybook/react';
+import Box from '@mui/material/Box';
+import dedent from 'ts-dedent';
+import {EzCard} from '../../../EzCard';
+import EzStackedBarChart from '../../EzStackedBarChart';
+import EzHeading from '../../../EzHeading';
+
+const meta: Meta<typeof EzStackedBarChart> = {
+  argTypes: {
+    displayData: {
```
noranda, line 11: We call this `data` in `EzBarChart`. Should these match?

packages/recipe/src/components/EzStackedBarChart/Implementations/CustomTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <g style={{pointerEvents: 'none', fontFamily, fontSize: 10}}>
+      <foreignObject width="150" height="400" x={props.x + 16} y={props.y}>
+        <div
+          style={{
+            backgroundColor: 'white',
+            borderRadius: 4,
+            boxShadow: '0px 16px 32px 0px rgba(0, 0, 0, 0.10), 0px 0px 4px 0px rgba(0, 0, 0, 0.04)',
+          }}
+        >
```
noranda, line 20: Should we use MUI components for this? We should also use design tokens where possible.

packages/recipe/src/components/EzStackedBarChart/Implementations/CustomTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+            }}
+          >
+            <div>{props.datum.xName}</div>
+            <div>{dependentAxisLabelFormatter(total)}</div>
+          </div>
+          <div style={{padding: '8px 16px 0px 16px'}}>
+            {props.data.map((datum, i) => {
+              const {barColor, label, data} = datum;
+              // eslint-disable-next-line react-hooks/rules-of-hooks
+              const color = useThemeColor(barColor);
```
noranda, line 36: Is there another way we can write this where we don't have to break this hooks rule?

packages/recipe/src/components/EzStackedBarChart/Implementations/EzStackedBarChartVictory.tsx
```diff
@@ -0,0 +1,142 @@
...
+    } = getVictoryChartStyles({
+      fontFamily,
+    });
+    const {legendStyles} = getVictoryLegendStyles({fontFamily});
+    const legendData = displayData.map(datum => {
+      return {
+        name: datum.label,
+        symbol: {
+          // eslint-disable-next-line react-hooks/rules-of-hooks
+          fill: useThemeColor(datum.barColor),
```
noranda, line 58: Can we pull this out so we don't have to break this hooks rule?

packages/recipe/src/components/EzStackedBarChart/Implementations/EzStackedBarChartVictory.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    data={displayData}
+                    dependentAxisLabelFormatter={dependentAxisLabelFormatter}
+                  />
+                }
+              />
+            }
+          >
+            {displayData.map(datum => {
+              // eslint-disable-next-line react-hooks/rules-of-hooks
+              const barFillColor = useThemeColor(datum.barColor);
```
noranda, line 120: Can we break this out?

packages/recipe/src/components/EzStackedBarChart/__tests__/EzStackedBarChart.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+    );
+
+    expect(getByText(`${MAX_DEPENDENT_VALUE}`)).toBeInTheDocument();
+    expect(getByText(`${MAX_INDEPENDENT_VALUE}`)).toBeInTheDocument();
+    expect(queryByText(`${MAX_DEPENDENT_VALUE + INCREMENT}`)).not.toBeInTheDocument();
+    expect(queryByText(`${MAX_INDEPENDENT_VALUE + 1}`)).not.toBeInTheDocument();
+  });
+});
+
+describe('EzBarChart', () => {
```
noranda, line 145:
```suggestion
describe('EzStackedBarChart', () => {
```

packages/recipe/src/components/index.ts
```diff
@@ -47,3 +47,4 @@ export {EzCard, EzCardSection, EzCardExpandable} from './EzCard';
 export {EzContent, EzHeader, EzItem, EzFooter, EzPreview} from './EzContent';
 export {EzPage, EzPageSection} from './EzPage';
 export {EzTimeline, EzTimelineEvent, EzTimelinePeriod} from './EzTimeline';
+export {default as EzStackedBarChart} from './EzStackedBarChart';
```
noranda, line 50: Can we sort this?

noranda (changes requested):
Nice job! I have a few requested changes and a clarification question below.

1. Clicking on the preview button in the Component Library breaks the app.
2. The playroom examples don't work.
3. See various other change requests in the code.

I realize that having more than one data set requires a separate implementation, but this component's arg types are nearly identical to `EzBarChart`. Is there a way we could combine them into a single component and render the correct implementation depending on number of data sets provided? I know this would mean a breaking change, but perhaps it is the correct thing to do here? (CC @alexanderlippner16)

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
danidewitt: it says "over multiple categories" and each dataset is a category in a bar chart

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
danidewitt: i'm actually using this, which is why it is exported

packages/recipe/src/components/EzStackedBarChart/Documentation/Stories/Default.stories.tsx
danidewitt: this is more than just data, it also includes color and label information for each set of data, so i'd argue it should be different

packages/recipe/src/components/EzStackedBarChart/Implementations/CustomTooltip.tsx
danidewitt: this was in my list for v2 due to deadlines, but i just did it

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.snapshot.tsx
danidewitt: is that true for our other components too? because i see a bunch of other components rendering the component

danidewitt: the rendering issue was just a case i hadn't accounted for, i think this is otherwise fine now.

danidewitt:
@noranda i've responded to all the comments and implemented relevant changes.

as for why it's a separate component, actually bar charts and stacked bar charts are not the same (as [noted](https://github.com/ezcater/recipe/pull/1076/files#diff-109006220a7809c181328d175acac0d23ade47a1002a43b417305cf55f4a2caeR16) in the documentation), even though they have similar (but different) props. see the difference explained [here](https://www.storytellingwithdata.com/blog/stacked-bars). specifcially: 
>  While this may seem obvious, a stacked bar chart is not the same as a [standard bar chart](https://www.storytellingwithdata.com/blog/2020/2/19/what-is-a-bar-chart). A traditional bar chart displays a single categorical variable in the form of bars. A stacked bar chart shows two categorical variables. The first (and primary) variable is shown along the entire length of the bar, and the second variable is represented as stacks within each categorical bar.

we're in the process of breaking apart our other components, like EzField, and i think we've seen the complexity and problems that arise when you smash multiple components into one, try to make the props "figure" things out, then need to pull the components apart.

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
noranda: It says "especially" which makes it seem like it's fine to use this component with a single category. Is that the case?

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.mdx
noranda: This is just the MDX though, no? In an case, not a big deal.

packages/recipe/src/components/EzStackedBarChart/Documentation/EzStackedBarChart.snapshot.tsx
noranda: If the component is highly interactive (like a chart) I would suggest we abstract the component so that we don't run into rendering issues, but so long as it works here, I'm fine with it.

noranda (changes requested):
The changes look good, except for one clarification and the playroom examples still don't work.

With regard to keeping this a separate component, I'd still like @alexanderlippner16 to weigh in, but otherwise I'm fine with it. While it's true that some components should be broken out (like `EzField`), this is because we need to mix and match inputs in groups. I don't think we need to do anything like that here, but I'll let Alexis make that call.

danidewitt:
@noranda what's wrong with the playroom examples? they're working for me:
Screenshot 2023-11-07 at 9 36 02 AM

noranda (approved):

danidewitt:
@noranda as for the separate component: we did discuss making it a separate component in meetings and everyone was 👍 with it. it's the highest priority request, if it's not what we want, we need to let our customer know it's going to be a longer wait.

in any case, i still stand by stacked bar charts being a separate kind of chart than bar charts, given the argument in the article i [linked above](https://www.storytellingwithdata.com/blog/stacked-bars). the props need to be different, the tooltip is different, the purpose is different. there are many variations on stacked bar charts that we might want in the future, and i can see bar charts becoming impossibly bloated and difficult to maintain if we don't separate them now. two examples that immediately come to mind: cluster bar charts and cluster stacked bar charts.

pgrippi (approved):

